PR #3, opened 2023-09-16
Add Google tests for MVP functionality

jbcoe:
Write code without knowing about implementation details.

May be enlightening, seems to be as some tests need to be disabled.


jbcoe:
Drop in coverage is due to disabled tests - will merge and address later

PR #5, opened 2023-09-16
Add minimal implementations of indirect and polymorphic with basic allocator support

jbcoe:
Constructors taking an allocator should be added in a later PR.

More allocator-focussed tests should be added to ensure that we're cleaning things up correctly (again a later PR).

### DIFF
--- a/indirect_with_allocators.h
+++ b/indirect_with_allocators.h
@@ -1,3 +1,5 @@
+#ifndef INDIRECT_WITH_ALLOCATORS_H
+#define INDIRECT_WITH_ALLOCATORS_H
 #include <memory>
 #include <utility>
 
@@ -98,3 +100,5 @@ struct std::hash<xyz::indirect<T>> {
     return std::hash<typename xyz::indirect<T>::value_type>{}(*key);
   }
 };
+
+#endif  // INDIRECT_WITH_ALLOCATORS_H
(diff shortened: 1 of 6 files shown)


jbcoe:
Code coverage drop is ok - some tests have disabled code when we're not using allocators.

indirect_with_allocators.h
```diff
@@ -0,0 +1,106 @@
+#ifndef XYZ_INDIRECT_WITH_ALLOCATORS_H
+#define XYZ_INDIRECT_WITH_ALLOCATORS_H
+#include <memory>
+#include <utility>
+
+namespace xyz {
+template <class T, class A = std::allocator<T>>
+class indirect {
+  T* p_;
+  [[no_unique_address]] A alloc_;
```
Twon, line 10: MSVC ignores `[[no_unique_address]]` but they have their own ``[[msvc::no_unique_address]]` which will be necessary here for MSVC to pass a test that `sizeof(indirect<T>) == sizeof(T*)`

indirect_with_allocators.h
```diff
@@ -0,0 +1,106 @@
+#ifndef XYZ_INDIRECT_WITH_ALLOCATORS_H
+#define XYZ_INDIRECT_WITH_ALLOCATORS_H
+#include <memory>
+#include <utility>
+
+namespace xyz {
+template <class T, class A = std::allocator<T>>
```
Twon, line 7: Presumably `indirect` should now have a `get_allocator` method like other container types?

Twon (approved):

PR #9, opened 2023-09-17
Avoid vtable for getting pointer from control block

jbcoe:
Merge this later.


Twon (approved):
Nice way to avoid the virtual dispatch cost of `get`

PR #12, opened 2023-09-18
Add exception safety and tests for exceptions

jbcoe:


Twon (approved):
All make sense and as expected.

PR #15, opened 2023-09-18
Make non-allocating member functions constexpr

jbcoe:


Twon (approved):
Does what it says on the tin...

PR #16, opened 2023-09-18
Use a single set of headers and tests

jbcoe:
* Delete old headers and tests.

* Update build targets.

* Rename tests and headers (possibly contentious)


jbcoe:
Coverage drop is probably to be expected given the amount of code we are deleting.

It might be worth getting coverage for new headers to 100% on a separate PR.

Twon (approved):
Makes sense to build against the same tests.

Twon:
> Coverage drop is probably to be expected given the amount of code we are deleting.
> 
> It might be worth getting coverage for new headers to 100% on a separate PR.

Yes, I am happy with that.  Let's get it merged for now.

Issue #25, opened 2023-09-21
Get code coverage up to 100% for headers


jbcoe:

jbcoe:
Addressed by #28

PR #28, opened 2023-09-22
Add coverage for polymorphic clone method and exceptions

jbcoe:


jbcoe:
Raises coverage of header files to 100%.

Coverage of tests themselves is less of a concern.

jbcoe:
closes #25

PR #31, opened 2023-09-23
Add missing assertions for valueless in swap

jbcoe:


jbcoe:
Trivial changes - will just merge these.

PR #38, opened 2023-09-24
Add support for comparison of indirect<T> with U

jbcoe:


jbcoe:
I can't see how to get concept requirements working without recursion. Save fixing this for later.

PR #43, opened 2023-09-24
Add wording for null state and value semantics

jbcoe:


jbcoe:
No code changes - just design notes.

PR #47, opened 2023-09-24
Constexprfy all methods

Twon:


jbcoe (commented):
Can we add tests to ensure that these functions can meaningfully be `constexpr`?

Ukilele:
> Can we add tests to ensure that these functions can meaningfully be `constexpr`?

I hope I understand your question correctly: You can have a look at the [test-suit of MSVC STL](https://github.com/microsoft/STL/blob/main/tests/std/tests/P0202R3_constexpr_algorithm_and_exchange/test.cpp#L627-L628). They write all their test-functions as constexpr. And then in their main function they call the function both from "runtime-context" and "compile-time-context":
```cpp
int main() {
    test();
    static_assert(test());
}
```
Unfortunately I am not sure yet how to combine this approach with Google Test.

jbcoe:
#65 should replace this. Tests pass on AppleClang with bazel but not with GCC. 

Will work on a local (Docker) repro.

@Ukilele input welcome if you're interested/able.

PR #48, opened 2023-09-24
Expand Small object optimisation for

jbcoe:


jbcoe:
Changes to the draft, not the code.

PR #51, opened 2023-09-25
Add allocator_arg_t to all constructors that take an allocator

jbcoe:


Twon (commented):
Be interested to understand the rational behind this from Nina Ranns

PR #52, opened 2023-09-25
Edits for Titus feedback

jbcoe:


Twon (approved):
I think that is useful clarification to add.

PR #58, opened 2023-09-28
    Reused the term control_block for a struct containing function po…

BengtGustafsson:
…inters to destroy and clone.

    polymorphic itself was changed to hold two pointers, one to the control block and one to the data itself. This avoids the double indirection of the original design and lets the conrtol block    be independent of the actual data pointer.

    A problem is when and how to initiaate the function pointers. I found three options, and implemented all using a ifdef. All three methods rely on  a constructor of control_block taking an in_place_type<U> which sets up the function pointers.

    1. Use the ctor to construct the instances of a variable template. This causes an issue with initialization order (at least if the same T, U, A combination is used in multiple TUs. init_control_block just returns an instance of the variable template.

    2. Replace the variable template with a singleton inside the init_control_block function. This is probably the most correct wat but costly due to the thread-safe single init overhead code.

    3.The init_control_block function initializes the variable template instance each time. The problem is that this may be slightly UB with multiple threads as they may write the same memory location simultaneously. But as they all write the same value this is ok in practice.

This implementation allows classes using polymorphic<T> to be defined with incomplete T as long as their ctors that construct the polymorphic<T> with some U have seen a definition of U (and thereby T) when defined. I added a "test case" for this which compiles code in an order that wouldn't work if this was not true. Maybe a fuller test with multiple TUs should be written just to make really sure that this is true.

### DIFF
--- a/polymorphic.h
+++ b/polymorphic.h
@@ -26,57 +26,89 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <memory>
 #include <utility>
 
+#define IMPL_OPTION 2
+
 namespace xyz {
 
 namespace detail {
+
 template <class T, class A>
 struct control_block {
-  T* p_;
-  virtual ~control_block() = default;
-  virtual void destroy(A& alloc) = 0;
-  virtual control_block<T, A>* clone(A& alloc) = 0;
+  control_block() {}
+  template<typename U> control_block(std::in_place_type_t<U>) {
+    destroy = [](A& alloc, T* src) {
+      using u_allocator = typename std::allocator_traits<A>::template rebind_alloc<U>;
+      using u_alloc_traits = std::allocator_traits<u_allocator>;
+
+      u_allocator u_alloc(alloc);
+      u_alloc_traits::destroy(u_alloc, src);
+      u_alloc_traits::deallocate(u_alloc, static_cast<U*>(src), 1);
+    };
+
+    clone = [](A& alloc, const T& src)->T* {
+      using u_allocator = typename std::allocator_traits<A>::template rebind_alloc<U>;
+      using u_alloc_traits = std::allocator_traits<u_allocator>;
+
+      u_allocator u_alloc(alloc);
+      auto* mem = u_alloc_traits::allocate(u_alloc, 1);
+      try {
+        u_alloc_traits::construct(u_alloc, mem, static_cast<const U&>(src));
+        return mem;
+      } catch (...) {
+        u_alloc_traits::deallocate(u_alloc, mem, 1);
+        throw;
+      }
+    };
+  }
+
+  void (*destroy)(A& alloc, T* src);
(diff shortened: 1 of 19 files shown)


polymorphic.h
```diff
@@ -132,99 +161,108 @@ class polymorphic {
...
   }
 
   constexpr bool valueless_after_move() const noexcept {
-    return cb_ == nullptr;
+    return p_ == nullptr;
   }
 
   constexpr void swap(polymorphic& other) noexcept {
-    assert(other.cb_ != nullptr);  // LCOV_EXCL_LINE
+    // NOTE: I don't think these asserts should be here. Even if one is valueless it should be swappable
```
jbcoe, line 245: I think that these asserts are correct. We require that move-from is the only way to get into a valueless state after which any access to the object other than to assign to it or destroy it is invalid.

polymorphic.h
```diff
@@ -26,81 +26,110 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+  static control_block<T, A> instance(std::in_place_type<U>);
+  return &instance;
+}
+
+#elif IMPL_OPTION == 3
+
+template<class T, class U, class A> inline control_block<T, A> s_control_block;
+
+// Set up the control block each time to make sure it is set early enough without the function-static variable overhead.
+// Thread safety is so-so as multiple threads may write the same values to the function pointers at the same time, which could be
```
jbcoe, line 93: Could we add some tests for this thread-safety concern? I'm not 100% sure I follow.

jbcoe:
If I've understood this branch correctly, I think it's mostly captured by #69

@BengtGustafsson are we happy to close?

jbcoe:
PR #79 adds an experimental folder. Could your alternative implementation go in there too once the PR is merged?

It would be nice to consolidate the tests so that all implementations use the same ones.

polymorphic.h
BengtGustafsson: I would argue that swap is just a double assign and should be allowed. The litmus test would be whether a variant in valueless_by_exception state is swappable.

polymorphic.h
BengtGustafsson: I don't think it is a concern in reality, but probably officially UB. Each implementation would have to decide if it is ok to asynchronously write the same value to the same address in multiple cores. I have a hard time thinking of a hardware where this could fail, but in the "abstract machine" I'm pretty sure it is UB.

This design is however pretty old by now. The design I would go for (as an implementer) is now in cloning_ptr.h which allows up- and downcasting as per your original design that I found flawed. I don't understand why we didn't come up with this solution at the time. Maybe because we didn't want to accept a fat pointer _or_ an extra indirection. I now believe that you need to have either a fat pointer (like shared_ptr) or an indirection level (as in your polymorphic.h) but if you have it you can also allow casting as for shared_ptr.

jbcoe:
I think we've assimilated all the things we want for Kona. Feel free to re-open if I've missed things.

PR #59, opened 2023-09-30
Use traits to determine allocator propagation

### LINKED ISSUE #66: Use allocator traits for allocator propagation

jbcoe:
Fixes #66

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -171,7 +171,8 @@ class indirect {
   }
 
   constexpr indirect& operator=(indirect&& other) noexcept(
-      allocator_traits::propagate_on_container_move_assignment::value) {
+      allocator_traits::propagate_on_container_move_assignment::value ||
+      allocator_traits::is_always_equal::value) {
     assert(other.p_ != nullptr);  // LCOV_EXCL_LINE
     if (this != &other) {
       reset();
--- a/indirect.h
+++ b/indirect.h
@@ -223,16 +223,21 @@ class indirect {
 
   constexpr allocator_type get_allocator() const noexcept { return alloc_; }
 
-  constexpr void swap(indirect& other) noexcept {
+  constexpr void swap(indirect& other) noexcept(
+      allocator_traits::propagate_on_container_swap::value ||
+      allocator_traits::is_always_equal::value) {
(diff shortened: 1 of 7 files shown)


jbcoe:
Complicated merge to do - I'll sort this

jbcoe:
I’ll resolve the merge conflicts once I’m home

indirect.h
```diff
@@ -140,31 +144,57 @@ class indirect {
...
-    } else {
-      indirect tmp(other);
-      this->swap(tmp);
     }
     return *this;
   }
 
-  constexpr indirect& operator=(indirect&& other) noexcept {
+  constexpr indirect& operator=(indirect&& other) noexcept(
+      allocator_traits::propagate_on_container_move_assignment::value) {
```
Twon, line 174: I belive this should also be conditional with or of allocator traits is_always_equal as is the case for std::vector swap https://eel.is/c++draft/vector.capacity
```suggestion
      allocator_traits::propagate_on_container_move_assignment::value || allocator_traits::is_always_equal::value) {
```

jbcoe: Agreed. Can you make the change?

Twon: Yes, just pushed up.  I'm working on the swap methods now

Twon: Interestingly it looks like no major STL implementations support the case where propagate_on_container_move_assignment is false or allocator are not equal:

Microsoft: https://github.com/microsoft/STL/blob/f362f7d7bea87bd199c927a89718e929ed89b187/stl/inc/vector#L3452

libstdc++: https://github.com/gcc-mirror/gcc/blob/9a82cca91dee190d845ee1681440bc69a07c6d49/libstdc%2B%2B-v3/include/bits/stl_vector.h#L1591

libc++: https://github.com/llvm/llvm-project/blob/5d35273a32d239b7407338e13ed71b59174d6536/libcxx/include/vector#L2017

Twon (approved):
The change match my expectation based on my understanding of this via: https://foonathan.net/2015/10/allocatorawarecontainer-propagation-pitfalls/

I've now extended support for POCS also.

PR #61, opened 2023-09-30
Call std::swap directly

### LINKED ISSUE #57: Make `swap` ADL-proof
In the implementation of `indirect` and `polymporphic` `swap` gets called 8 times. Sometimes the member functions `indirect::swap` and `polymorphic::swap` (e.g. `swap(tmp);`) get called. And sometimes `std::swap` (https://eel.is/c++draft/utility.swap#lib:swap) on two pointers (e.g. `swap(p_, other.p_)` with `using std::swap` in the line above) gets called. The point is that `swap` gets called unqualified, i.e. it is subject to ADL. But we already know which overloads of `swap` we want to call and don't need/want different overloads to be picked by ADL. While there is no advantage in calling `swap` unqualified, there might be a disadvantage in doing so: I came up with an example, even though it is very contrived:
```cpp
template<class T>
struct Holder { T value; };

class Incomplete;

template<class T>
struct X {};

using Indirect = xyz::indirect<X<Holder<Incomplete>>>;

void TestSwap(Indirect& a, Indirect& b)
{
    a.swap(b);
}
```
The snippet does not compile, as `a.swap(b)` makes an unqualified call to `swap` [here](https://github.com/jbcoe/value_types/blob/main/indirect.h#L184). The arguments  (`p_` and `other.p_`) are of type `X<Holder<Incomplete>>*`. To perform ADL, the compiler has to instantiate several types, including `Holder<Incomplete>`. If you want to protect against this issue, I suggest to fully qualify all calls to `swap`.

If you want to have a more in depth discussion of the topic, I can recommend [ADL can interfere even with uglified names](https://quuxplusone.github.io/blog/2019/09/26/uglification-doesnt-stop-adl/) and [What is the std::swap two-step?](https://quuxplusone.github.io/blog/2020/07/11/the-std-swap-two-step/).

jbcoe:
fixes #57

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -125,15 +125,13 @@ class indirect {
   indirect(indirect&& other) noexcept
       : p_(nullptr), alloc_(std::move(other.alloc_)) {
     assert(other.p_ != nullptr);  // LCOV_EXCL_LINE
-    using std::swap;
-    swap(p_, other.p_);
+    std::swap(p_, other.p_);
   }
 
   indirect(std::allocator_arg_t, const A& alloc, indirect&& other) noexcept
       : p_(nullptr), alloc_(alloc) {
     assert(other.p_ != nullptr);  // LCOV_EXCL_LINE
-    using std::swap;
-    swap(p_, other.p_);
+    std::swap(p_, other.p_);
   }
 
   ~indirect() { reset(); }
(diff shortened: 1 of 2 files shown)


Ukilele (commented):
In indirect.h in [line 146](https://github.com/jbcoe/value_types/pull/61/files#diff-01332f5efea881b38f49805d6919e3525724aaadaf53ba1a3cb056d24850aae1L146) `swap(tmp);` also still invokes ADL and should be replaced by `this->swap(tmp);` to avoid ADL.

PR #64, opened 2023-09-30
Turn off inlining for code coverage runs

jbcoe:


jbcoe:
Disabling inlining made no odds to code coverage.

PR #69, opened 2023-10-03
[Experiment] Attempt to write an explicit vtable to improve polymorphic benchmark performance

jbcoe:


polymorphic.h
```diff
@@ -31,45 +31,64 @@ namespace xyz {
...
+
+template <class T, class A>
+struct fake_control_block : public control_block<T, A> {
+  fake_control_block() = delete;
+  std::byte data_;
 };
 
 template <class T, class U, class A>
 class direct_control_block : public control_block<T, A> {
   U u_;
```
BengtGustafsson, line 60: I don't think this will work, if I understand how you want it to work: If U has an alignment higher than sizeof(void*) the offset to fake_control_block::data_ will not be the same as the offset to u_ here.

polymorphic.h
```diff
@@ -183,24 +205,28 @@ class polymorphic {
     return *this;
   }
 
+  constexpr T* get_pointer() const noexcept {
+    return (T*)&reinterpret_cast<detail::fake_control_block<T, A>*>(cb_)->data_;
```
BengtGustafsson, line 209: This type of wild casting won't work if T is the second base of U, and less so if it is a virtual base. I think you'd need a third dispatch point for casting, which is unacceptable performance wise*. So I think the pointer has to be fat, with the data pointer pointing to a T. This is what led me to just allocating the U on the heap and then point to a _statically allocated_ control block with the cb_ pointer.

* An alternative could be to store a U->T offset in the control block, but it still requires an access in the control block just to calculate the address of its u_.

I don't think a two pointer fatness is that bad, its the same as shared_ptr already has.

I'm working on an idea to get my PR's implementation to handle virtual bases, which it doesn't right now.

PR #70, opened 2023-10-03
Make indirect-assign conditionally not allocate

jbcoe:
This is a significant design change as it prevents `indirect` from holding a object of a derived type of a non-copyable base class (if the base class were also non-movable then using `indirect` would be fine). Given the check on movability, we can still support moving `indirect<T>` when `T` is a non-movable type (like a mutex).

This change avoid allocation and makes `indirect` cheaper, which as a low-cost abstraction is the direction we want.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -445,9 +445,13 @@ indirect& operator=(const indirect& other);
 
 * _Preconditions_: `other` is not valueless.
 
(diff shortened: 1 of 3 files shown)


BengtGustafsson (approved):
Nothing to remark.

Twon (approved):
Great, I agree this is a worthwhile improvement and add extra benefits to using indirect.

PR #73, opened 2023-10-07
Remove unnecessary lcov exclusion noise

jbcoe:


jbcoe:
Exclusions are needed. Closing PR.

PR #79, opened 2023-10-11
Added experimental inline vtable for polymorphic

jbcoe:
Add an experimental implementation of polymorphic to try and get better performance. Set up polymorphic header so that alternative implementations can be selected with a preprocessor define. Add test and benchmark targets for the new experimental implementation (tests pass) in bazel.

TODO: Add CMake targets. Update: Done


CMakeLists.txt
```diff
@@ -130,26 +142,61 @@ if (${CPP_VALUE_TYPES_IS_NOT_SUBPROJECT})
...
         )
 
+        add_executable(polymorphic_inline_vtable_test "")
+        target_sources(polymorphic_inline_vtable_test
+            PRIVATE
+                polymorphic_test.cc
+        )
+        target_link_libraries(polymorphic_inline_vtable_test
+            PRIVATE
+            polymorphic_inline_vtable
```
Twon, line 159:
```suggestion
                polymorphic_inline_vtable
```

Twon (commented):
Looks like it could use updating to include the progagate_on_container_swap check to. But otherwise looks good

Twon (approved):

jbcoe:
Does polymorphic.h non-member swap need progagate_on_container_swap too?

PR #89, opened 2023-10-19
A feature test macro to spec

Twon:


DRAFT.md
```diff
@@ -1013,6 +1013,13 @@ constexpr polymorphic<T, Alloc>& operator->() noexcept;
 
 Otherwise, the interface of the specialization is as defined in [optional].
 
+## Feature-test Macro
+Add a new feature-test macro:
+
+```c++
+#define __cpp_lib_value_types 2023XXL
```
jbcoe, line 1020: Do we want two, one for indirect, one for polymorphic?

Twon: I was unsure here, but happy to go with 2.

jbcoe: There are two class templates so we should probably give implementers flexibility in supporting just one.

PR #90, opened 2023-10-19
Extend indirect to work with std::format

Twon:
Initial draft, will push up tests and add to the spec shortly.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -720,7 +720,6 @@ Otherwise, the interface of the specialization is as defined in [optional].
 ```c++
 // [indirect.fmt]
 template <class T, class Alloc, class charT>
-  requires { std::formatter<T, char>{}; };
 struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
   constexpr typename FormatContext::iterator parse(format_parse_context& ctx);
 
@@ -731,7 +730,9 @@ struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
 ```
 
 Specialization of `std::formatter<indirect<T, Alloc>, charT>` when the underlying
-T support specialisation of `std::formatter<T, charT>`.
+T supports specialisation of `std::formatter<T, charT>`.
+
+* Preconditions: The specialization formatter<T, charT> meets the Formatter requirements.
 
 ## Feature-test Macro
 Add a new feature-test macro:
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -715,6 +715,24 @@ constexpr indirect<T, Alloc>& operator->() noexcept;
 
 Otherwise, the interface of the specialization is as defined in [optional].
 
+#### X.Y.13 Formatter support [indirect.fmt]
+
+```c++
+// [indirect.fmt]
+template <class T, class Alloc, class charT>
+  requires { std::formatter<T, char>{}; };
+struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
+  constexpr typename FormatContext::iterator parse(format_parse_context& ctx);
+
+  template<class FormatContext>
+  typename FormatContext::iterator format(
+    indirect<T, Alloc> const& value, FormatContext& ctx) const;
+};
+```
+
+Specialization of `std::formatter<indirect<T, Alloc>, charT>` when the underlying
+T support specialisation of `std::formatter<T, charT>`.
+
 ### X.Z Class template polymorphic [polymorphic]
 
 #### X.Z.1 Class template polymorphic general [polymorphic.general]
--- a/indirect.h
+++ b/indirect.h
@@ -26,6 +26,10 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <memory>
 #include <utility>
 
+#if __has_include(<format>)
+#include <format>
+#endif
+
 namespace xyz {
 
 template <class T, class A>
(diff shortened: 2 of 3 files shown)


indirect.h
```diff
@@ -343,4 +347,29 @@ struct std::hash<xyz::indirect<T, Alloc>> {
...
+    return std::begin(ctx);
+  }
+
+  auto format(xyz::indirect<T, Alloc> const& value, format_context& ctx) const
+      -> format_context::iterator {
+    using namespace std::literals;
+    return std::format_to(ctx.out(), "{}", *value);
+  }
+};
+#endif
```
jbcoe, line 373: Nit: Missing `// __cpp_lib_format >= 201907L`

indirect.h
```diff
@@ -343,4 +347,29 @@ struct std::hash<xyz::indirect<T, Alloc>> {
...
+struct std::formatter<xyz::indirect<T, Alloc>> {
+  constexpr auto parse(format_parse_context& ctx)
+      -> format_parse_context::iterator {
+    return std::begin(ctx);
+  }
+
+  auto format(xyz::indirect<T, Alloc> const& value, format_context& ctx) const
+      -> format_context::iterator {
+    using namespace std::literals;
+    return std::format_to(ctx.out(), "{}", *value);
```
Twon, line 370: Presumably propagation of `format_context` is required here... needs testing once implemented.

Twon: This is working now

DRAFT.md
```diff
@@ -715,6 +715,24 @@ constexpr indirect<T, Alloc>& operator->() noexcept;
 
 Otherwise, the interface of the specialization is as defined in [optional].
 
+#### X.Y.13 Formatter support [indirect.fmt]
+
+```c++
+// [indirect.fmt]
+template <class T, class Alloc, class charT>
+  requires { std::formatter<T, char>{}; };
```
Twon, line 723: @jbcoe does this look like how a requirement should be specified?  Note in the implementation I created a concept for this, but here I've used an inline requires of the same check.

DRAFT.md
```diff
@@ -715,6 +715,24 @@ constexpr indirect<T, Alloc>& operator->() noexcept;
...
+struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
+  constexpr typename FormatContext::iterator parse(format_parse_context& ctx);
+
+  template<class FormatContext>
+  typename FormatContext::iterator format(
+    indirect<T, Alloc> const& value, FormatContext& ctx) const;
+};
+```
+
+Specialization of `std::formatter<indirect<T, Alloc>, charT>` when the underlying
```
Twon, line 733: @jbcoe let me know if the wording can be improved here.

jbcoe: Maybe “T supports” (missing s) but otherwise looks great.

DRAFT.md
jbcoe: I don’t think that the standard uses concept requirements in code. Just text in a _Requires_ clause

indirect_test.cc
```diff
@@ -647,4 +647,23 @@ TEST(IndirectTest, HashCustomAllocator) {
...
+#if (__cpp_lib_format >= 201907L)
+
+TEST(IndirectTest, FormatNativeTypesDefaultFormatting) {
+  EXPECT_EQ(std::format("{}", xyz::indirect<bool>(std::in_place, true)), "true");
+  EXPECT_EQ(std::format("{}", xyz::indirect<int>(std::in_place, 100)), "100");
+  EXPECT_EQ(std::format("{}", xyz::indirect<float>(std::in_place, 50.0f)), "50");
+  EXPECT_EQ(std::format("{}", xyz::indirect<double>(std::in_place, 25.0)), "25");
+}
+
+TEST(IndirectTest, FormatNativeTypesPropagateFormatting) {
```
jbcoe, line 660: Nice tests!

DRAFT.md
Twon: Ok, added thanks.

DRAFT.md
Twon: I've added this as a precondition that "The specialization formatter<T, charT> meets the Formatter requirements.".  This requirement is mentioned in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1636r2.pdf

Twon:
Hopefully, that's all the issues addressed now.

jbcoe (approved):

DRAFT.md
```diff
@@ -715,6 +715,25 @@ constexpr indirect<T, Alloc>& operator->() noexcept;
 
 Otherwise, the interface of the specialization is as defined in [optional].
 
+#### X.Y.13 Formatter support [indirect.fmt]
+
+```c++
+// [indirect.fmt]
+template <class T, class Alloc, class charT>
+struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
+  constexpr typename FormatContext::iterator parse(format_parse_context& ctx);
```
Ukilele, line 724: I only have limited experience with `std::format`, but I think there are two issues in this line.
1. `FormatContext` gets used but I don't see where it is coming from
2. You support any `class charT`, but `parse` only takes a `format_parse_context` (which is an alias to `basic_format_parse_context<char>`.
Maybe the correct syntax would be: 
```c++
constexpr typename basic_format_context<charT>::iterator parse(basic_format_parse_context<charT>& ctx);
```
or if you take the `ParseContext` as template argument (as e.g. the [range_formatter](http://eel.is/c++draft/format.range#formatter) does):
```c++
template<class ParseContext>
constexpr typename ParseContext::iterator parse(ParseContext& ctx);
```
As your `format`-function takes the `FormatContext` as template argument, I guess it's consistent to also let the `parse`-function take its `ParseContext` as template argument, right?

indirect.h
```diff
@@ -343,4 +347,30 @@ struct std::hash<xyz::indirect<T, Alloc>> {
   }
 };
 
+#if (__cpp_lib_format >= 201907L)
+
+namespace xyz {
+
+template <class T>
+concept is_formattable = requires(T t) { std::formatter<T, char>{}; };
```
Ukilele, line 355: Since C++23 we have the concept [std::formattable<T, charT>](http://eel.is/c++draft/format.formattable#concept:formattable) which does exactly this. But I am not sure which standard libraries already implement this.

DRAFT.md
```diff
@@ -715,6 +715,25 @@ constexpr indirect<T, Alloc>& operator->() noexcept;
 
 Otherwise, the interface of the specialization is as defined in [optional].
 
+#### X.Y.13 Formatter support [indirect.fmt]
+
+```c++
+// [indirect.fmt]
+template <class T, class Alloc, class charT>
+struct std::formatter<indirect<T, Alloc>, charT> : std::formatter<T> {
```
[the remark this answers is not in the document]
jbcoe: Fixed in #92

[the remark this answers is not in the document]
Twon: Thanks, good catch

indirect.h
Twon: Looks like this is unsupported in Visual Studio yet, but is supported in GCC 13 onward and Clang 17 onwards.

DRAFT.md
Twon: Thanks, very limited experience with `std::format` yet so appreciate the feedback!

PR #92, opened 2023-10-20
Fixes from Ukilele for fmt

jbcoe:


indirect.h
```diff
@@ -351,13 +351,13 @@ struct std::hash<xyz::indirect<T, Alloc>> {
...
+template <class T, class charT>
+concept is_formattable = requires(T t) { std::formatter<T, charT>{}; };
     
-}
+} // namespace xyz
 
 template <class T, class Alloc, class charT>
-  requires xyz::is_formattable<T>
+  requires xyz::is_formattable<T, charT>
 struct std::formatter<xyz::indirect<T, Alloc>, charT> : std::formatter<T> {
```
Ukilele, line 361: Here I think is the last place where we need to replace `std::formatter<T>` by `std::formatter<T, charT>`.
Besides that looks good 👍

indirect.h
```diff
@@ -351,26 +351,26 @@ struct std::hash<xyz::indirect<T, Alloc>> {
...
 template <class T, class Alloc, class charT>
-  requires xyz::is_formattable<T>
-struct std::formatter<xyz::indirect<T, Alloc>, charT> : std::formatter<T> {
-  constexpr auto parse(format_parse_context& ctx)
-      -> format_parse_context::iterator {
+  requires xyz::is_formattable<T, charT>
+struct std::formatter<xyz::indirect<T, Alloc>, charT> : std::formatter<T, charT> {
+  template <class ParseContext>
+  constexpr auto parse(ParseContext& ctx) -> typename ParseContext::iterator {
     return std::formatter<T>::parse(ctx);
```
Ukilele, line 364: I am sorry, I am not very careful, it's weekend 🙂 
There are two more places, where `std::formatter<T>` needs to be replaced by `std::formatter<T, charT>`. Here and..

indirect.h
```diff
@@ -351,26 +351,26 @@ struct std::hash<xyz::indirect<T, Alloc>> {
...
   }
 
-  template<class FormatContext>
-  auto format(xyz::indirect<T, Alloc> const& value, FormatContext& ctx) const
-      -> typename FormatContext::iterator {
+  template <class FormatContext>
+  auto format(xyz::indirect<T, Alloc> const& value, FormatContext& ctx) const ->
+      typename FormatContext::iterator {
     return std::formatter<T>::format(*value, ctx);
```
Ukilele, line 370: ...here.

jbcoe: Well spotted. I think we've got them all now. 

Thanks as always for your keen eye and engagement.

indirect.h
```diff
@@ -351,26 +351,26 @@ struct std::hash<xyz::indirect<T, Alloc>> {
 
 namespace xyz {
 
-template <class T>
-concept is_formattable = requires(T t) { std::formatter<T, char>{}; };
-    
-}
+template <class T, class charT>
+concept is_formattable = requires(T t) { std::formatter<T, charT>{}; };
```
Twon, line 355: Could be:

```
#if (__cpp_lib_format_ranges >= 202207L)

template <class T, class charT>
concept formattable = std::formattable<T, charT>;

#else

template <class T, class charT>
concept formattable = requires { std::formatter<T, charT>{}; };

#endif  // (__cpp_lib_format_ranges >= 202207L)
```

I've tested this on Visual Studio and Clang 17.  Even though the [paper that introduced `std::formattable` used the `__cpp_lib_format` feature test flag](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2286r8.html#feature-test-macro) it looks like [a subsequent LWG issues was raised to change this to `__cpp_lib_format_ranges` due to to many concurrent papers bumping this feature macro](https://cplusplus.github.io/LWG/issue3750)

Twon (approved):
As mentioned, could use `std::formattable` when supported by the compile.  Otherwise looks good.

indirect.h
jbcoe: Lets merge and fix later where needed.

PR #97, opened 2023-10-27
Simplify requires expression

Ukilele:
`std::derived_from<U, T>` already returns true if `T` and `U` are the same type.


jbcoe:
Nicely spotted. Thanks for this.

PR #99, opened 2023-10-29
Remove specialisations of uses_allocator

jbcoe:
@Ukilele I think this does what you suggested.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -346,9 +346,6 @@ class indirect {
 };
 
 template <class T, class Alloc>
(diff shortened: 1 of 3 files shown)


Twon (approved):
Agreed, I've looked into this, and my understanding of 'std::uses_allocator' is its for type supporting allocator constructors but without the 'allocator_type' typedef. Types supporting  'allocator_type' are automatically detected as 'uses_allocator'.

Ukilele:
Yes, that's also my understanding.
The changes look good to me.

PR #100, opened 2023-10-29
Add licenses to test cc files

jbcoe:


jbcoe:
No changes outside of comment blocks. Will merge as-is if (when) tests pass.

PR #103, opened 2023-11-01
Dev twon allocator pointer

Twon:


Twon:
This could use tests.  Something like https://www.boost.org/doc/libs/1_83_0/doc/html/interprocess/offset_ptr.html would be idea to demonstrate this works.

jbcoe (approved):

jbcoe:
Testing this is non-trivial so I'd defer for now.

Ukilele (commented):
At least line 58 `T* mem = allocator_traits::allocate(alloc_, 1);` will fail to compile as we still expect a `T*` and not a `allocator_traits::pointer`.

Twon:
> Testing this is non-trivial so I'd defer for now.

I'm going to add a branch using Conan to fetch boost so that I can at least test this again `boost::interprocess::offset_ptr` to flush out all of the issues like the one @Ukilele has raised.  We won't check in these changes, but we can at least verify we've updated all of the places that should be using `pointer`

jbcoe:
Does this PR need accompanying wording changes? If so we need these very soon

Twon:
> Does this PR need accompanying wording changes? If so, we need these very soon.

I believe the wording in the draft is now updated. Please take a look and make sure you are happy with it.  We should get this checked in but this still needs further testing against boost interprocess allocators there are compiler errors to resolve in the test branch.

DRAFT.md
```diff
@@ -760,6 +762,8 @@ class polymorphic {
  public:
   using value_type = T;
   using allocator_type = Allocator;
+  using pointer       = typename allocator_traits<Allocator>::pointer;
```
jbcoe, line 765: I'd adjust the odd spacing.

DRAFT.md
```diff
@@ -272,6 +272,8 @@ class indirect {
...
```
jbcoe, line 275: I'd adjust the odd spacing.

DRAFT.md
```diff
@@ -1029,7 +1033,7 @@ Otherwise, the interface of the specialization is as defined in [optional].
 Add a new feature-test macro:
 
 ```c++
-#define __cpp_lib_value_types 2023XXL
+#define __cpp_lib_polymorphic 2023XXL
```
jbcoe, line 1036: Nicely spotted!

jbcoe (approved):
Great work. Approved with minor tweaks to whitespace.

jbcoe (approved):

PR #107, opened 2023-11-04
Polymorphic SBO

### LINKED ISSUE #102: Question about who no small object optimization?
This is just a question, sorry that I do not know where else to ask this.

I was skimming over the `polymorphic_value` papers and noticed that in the [original paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0201r5.html) small object allocation is encouraged, while in the [new paper](https://github.com/jbcoe/value_types/blob/main/DRAFT.md) it is expressly called out as not supported. I would just like to better understand the reason for the change? Was this due to something learned from the implementation, or perhaps a reconsideration of the tradeoffs involved?

The new paper does not provide much explanation, other than saying that a small object implementation would need to be a different type (true, but does not explain why that type was not chosen for the implementation).

jbcoe:
Add an experimental SBO implementation for polymorphic. 

We don't add consteval checks for this experimental variant because the implementation is reliant on variant::emplace being constexpr which did not happen until https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2231r1.html

There remains a compile issue with PMR allocators in the SBO implementation of polymorphic - this is left unresolved to be tackled in a follow up PR. This issue is captured in #112 

Fixes #102


jbcoe:
SBO benchmark results are encouraging but objects used in benchmark are perhaps unreasonably small to reflect real-world use.

```
MacBook Pro M2 Max

Run on (12 X 24 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x12)
Load Average: 1.16, 1.61, 1.62
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
Polymorphic_BM_Copy_RawPtr                          14.8 ns         14.8 ns     46788629
Polymorphic_BM_Copy_UniquePtr                       18.1 ns         18.1 ns     37696831
Polymorphic_BM_Copy_Polymorphic                     2.15 ns         2.15 ns    324688878
Polymorphic_BM_VectorCopy_RawPointer            81679219 ns     81678250 ns            8
Polymorphic_BM_VectorCopy_UniquePointer         39403380 ns     39403444 ns           18
Polymorphic_BM_VectorCopy_Polymorphic            9273768 ns      9273548 ns           73
Polymorphic_BM_ArrayCopy_RawPointer                41473 ns        41473 ns        16848
Polymorphic_BM_ArrayCopy_UniquePointer             37812 ns        37812 ns        18327
Polymorphic_BM_ArrayCopy_Polymorphic                4911 ns         4911 ns       142939
Polymorphic_BM_VectorAccumulate_RawPointer       1653403 ns      1653369 ns          420
Polymorphic_BM_VectorAccumulate_UniquePointer    1653077 ns      1653024 ns          423
Polymorphic_BM_VectorAccumulate_Polymorphic      3321299 ns      3321219 ns          210
```

DRAFT.md
```diff
@@ -1173,25 +1173,30 @@ outside of tests.
...
+
+It is possible to implement `polymorphic` with a small buffer optimisation,
+similar to that used in `std::function`. This would allow `polymorphic` to store
+small objects without allocating memory. Like `std::function`, the size of the
+small buffer is left to be specified by the implementation. Allowing a small
+buffer optimization means that constructors and potentially allocating member
+functions of polymorphic cannot be marked `constexpr`.
+
+The authors are sceptical of the value of a small buffer optimisation for
+objects from a type hierarchy. If the buffer is too small then all instances of
```
nbx8, line 1186: Minor edit (that -> than, break sentence in two):

If the buffer is too small, all instances of `polymorphic` will be larger than needed. This because they will allocate heap in addition to having memory from the buffer as part of the object size.

DRAFT.md
```diff
@@ -1173,25 +1173,30 @@ outside of tests.
...
+`polymorphic` will be larger that needed as they will allocate heap in addition
+to having the memory from the buffer as part of the object size. If the buffer
```
nbx8, line 1188: Minor edit (add comma, remove then): 

If the buffer is too big, polymorphic objects will be larger than necessary, potentially
introducing the need for `indirect<polymorphic<T>>`.

DRAFT.md
```diff
@@ -1173,25 +1173,30 @@ outside of tests.
...
+is too big then polymorphic objects will be larger than necessary, potentially
+introducing the need for `indirect<polymorphic<T>>`.
+
+With usage experience implementers will be able to determine if a small buffer
```
nbx8: Minor edit (add comma, experience -> experienced): 

With usage, experienced implementers will be able to determine if a small buffer optimisation is worthwhile.

jbcoe: Should be "With usage experience, implementers will be ...".

Fixed.

jbcoe:
Improved owned object access in updated impl (caches pointer avoiding need for indirection)

```
Run on (12 X 24 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x12)
Load Average: 2.09, 1.83, 1.67
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
Polymorphic_BM_Copy_RawPtr                          14.9 ns         14.9 ns     47290587
Polymorphic_BM_Copy_UniquePtr                       18.8 ns         18.8 ns     38362050
Polymorphic_BM_Copy_Polymorphic                     3.78 ns         3.78 ns    183709194
Polymorphic_BM_VectorCopy_RawPointer            75565417 ns     75565444 ns            9
Polymorphic_BM_VectorCopy_UniquePointer         39248468 ns     39248500 ns           18
Polymorphic_BM_VectorCopy_Polymorphic           12540995 ns     12540836 ns           55
Polymorphic_BM_ArrayCopy_RawPointer                41567 ns        41566 ns        16747
Polymorphic_BM_ArrayCopy_UniquePointer             37818 ns        37818 ns        18369
Polymorphic_BM_ArrayCopy_Polymorphic                6379 ns         6379 ns       106812
Polymorphic_BM_VectorAccumulate_RawPointer       1652118 ns      1652121 ns          423
Polymorphic_BM_VectorAccumulate_UniquePointer    1656900 ns      1656884 ns          422
Polymorphic_BM_VectorAccumulate_Polymorphic      1679343 ns      1679306 ns          412
```

DRAFT.md
```diff
@@ -1173,25 +1173,30 @@ outside of tests.
...
-N>` similar to `llvm::SmallVector<T, N>`, but we are not proposing its addition
-here.
+### Small Buffer Optimisation
+
+It is possible to implement `polymorphic` with a small buffer optimisation,
+similar to that used in `std::function`. This would allow `polymorphic` to store
+small objects without allocating memory. Like `std::function`, the size of the
+small buffer is left to be specified by the implementation. Allowing a small
+buffer optimization means that constructors and potentially allocating member
+functions of polymorphic cannot be marked `constexpr`.
```
Twon, line 1183: As discussed, we do not believe this to be an inherent problem with the language not allowing `constexpr` here but rather a result of the implementation using variant. 

The following paper added support for `constexpr` variants in C++23 but yet to be implemented by all compilers: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2231r1.html

DRAFT.md
```diff
@@ -1173,25 +1173,39 @@ outside of tests.
...
+small buffer is left to be specified by the implementation.
+
+The authors are sceptical of the value of a small buffer optimisation for
+objects from a type hierarchy. If the buffer is too small, all instances of
+`polymorphic` will be larger than needed. This is because they will allocate
+heap in addition to having the memory from the (empty) buffer as part of the
+object size. If the buffer is too big, `polymorphic` objects will be larger than
+necessary, potentially introducing the need for `indirect<polymorphic<T>>`.
+
+We could add another, non-type, template argument to `polymorphic` to specify
```
nbx8, line 1190: Minor edit (remove another):

We could add a non-type, template argument to `polymorphic` to specify
the size of the small buffer:

DRAFT.md
```diff
@@ -1173,25 +1173,39 @@ outside of tests.
...
+the size of the small buffer:
+
+```c++
+template <typename T, typename Alloc, size_t BufferSize>
+class polymorphic;
+```
+
+We opt not to provide an additional template argument to maintain consistency
```
nbx8, line 1198: Minor rewording: 

However we opt not to do this to maintain consistency with other standard library types.

DRAFT.md
```diff
@@ -1173,25 +1173,39 @@ outside of tests.
...
+with other standard library types. Both `std::function`. and `std::string` leave
```
nbx8, line 1199: Minor type (remove fullstop):

Both `std::function` and `std::string` leave the buffer size as an implmentation detail.

DRAFT.md
```diff
@@ -1173,25 +1173,39 @@ outside of tests.
...
+the buffer size as an implmentation detail. Adding an additional template
```
nbx8, line 1200: Minor edit (replace adding):

Including an additional template argument in a later revision of the standard would be a breaking change.

DRAFT.md
jbcoe: I've rewritten this section now, please re-review.

Twon (approved):
Looking good

PR #109, opened 2023-11-04
Fix for test allocators in GCC13

jbcoe:


jbcoe:
Merging this as the compile error it fixes is blocking tests on my Linux VM.

PR #113, opened 2023-11-05
Fixes for incomplete type support

jbcoe:
Add tests for compiling with incomplete types.

Update Draft and headers: some requires clauses need to be static_asserts to support incomplete types.

See: https://timsong-cpp.github.io/cppwp/n4861/structure for the difference between _Constraints_ and _Mandates_.

Thanks to Bengt Gustaffson for pointing out the need for these changes.

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -58,9 +58,7 @@ class indirect {
   using value_type = T;
   using allocator_type = A;
 
-  constexpr indirect()
-    requires std::default_initializable<T>
-  {
+  constexpr indirect() {
     T* mem = allocator_traits::allocate(alloc_, 1);
     try {
       allocator_traits::construct(alloc_, mem);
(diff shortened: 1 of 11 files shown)


indirect.h
```diff
@@ -58,9 +59,8 @@ class indirect {
   using value_type = T;
   using allocator_type = A;
 
-  constexpr indirect()
-    requires std::default_initializable<T>
-  {
+  constexpr indirect() {
+    static_assert(std::is_default_constructible_v<T>);
```
Twon, line 63: Why static asserts over requirements?

Twon (approved):
Minor query around static assert otherwise looks good

indirect.h
jbcoe: Requirements need complete types.

PR #117, opened 2023-11-06
[EXPERIMENT] Create an implementation and tests for unique_polymorphic

jbcoe:
I'm not at all sure that we want this, but it's possible to do.

DO NOT MERGE for now.


jbcoe:
@Ukilele if you fancy adding wording for this type, I think it’s a copy and paste job. Not sure there’s any real call for the type though.

jbcoe:
Closing PR for now to reduce noise

PR #119, opened 2023-11-06
Experiments with SBO and alignment

jbcoe:
I've widened the template arguments on the SBO polymorphic implementation so that we can oplay with buffer size and alignment. I don't think we want to ship anything like this, it leaves too much adjustment for a user to master.


jbcoe:
This PR is noise for now. Closing.

PR #120, opened 2023-11-07
Use clang17 on Ubuntu for CI

jbcoe:
Add CI for clang17 on Ubuntu. Clang17 seems new enough to work with C++20.


Twon:
CMake github action workflow is no longer running due to these changes.  As such I've added a PR for pre-checks which add an integrity checks workflow that should validate and catch issues in github action definitions: https://github.com/jbcoe/value_types/pull/126

Also to add Clang17 this will need to be installed manually as it's not supported by setup-clang: https://github.com/egor-tensin/setup-clang.  I will update this shortly to add support for it.

Twon (approved):
Looks like Clang-17 is working, and the https://github.com/egor-tensin/setup-clang action is simply out of date in not capturing this compiler is now supported

PR #121, opened 2023-11-07
Add draft presentation for Kona ISO standards meeting

jbcoe:


presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+| Idiom | Type |
+|---|---|
+| Variable-count, multiple, nullable instances of a type `T`| `std::vector<std::optional<T>>`|
+| Key-accessed, instances of one of a closed-set of types `Ts...`| `std::map<Key, std::variant<Ts...>>`|
+
+---
+
+# Incomplete types
+
+We may want a composite to contain an instance of an incomplete type, either directly or though the use of existing vocabulary types.
```
Ukilele, line 130: Typo (I think): "directly or through"

Ukilele (changes requested):
I did a review until chapter https://github.com/jbcoe/value_types/blob/dev-jbcoe-kona-slides-2023/presentations/vocabulary_types_2023_11_8.md#incomplete-types-using-pointers so far. I really like the presentation! Only minor typos that I found.

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+    // ...
+}
+```
+
+Again the compiler-generated copy constructor and assignment operator will give rise to multiple composite objects that share the same components, _but_ the shared components 
+are immutable. 
+
+If our hierarchy of composites is const-correct, immutability will hide the fact that data is shared.
+
+We can't call non-const qualified member function on our component `ptr_`.
```
Ukilele, line 208: Tiny nitpick: You can call non-const qualified member function**s** on `ptr_`. What you can't do is call non-const qualified member functions on `*ptr_` (the pointee).

jbcoe: Thanks, this is no nitpick but a significant mistake - well spotted!

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+## [P3019r0](https://wg21.link/p3019r0)
+
+### Kona Hawaii 2023-11-8
+
+
+---
+
+# Introduction
+
+We propose adding two new class templates to the C++ Standard Library.
```
nbx8, line 24: We propose adding two new class templates to the C++ Standard Library:

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+class polymorphic;
+```
+
+These fill an existing gap in the suite of existing standard library vocabulary types.
+
+---
+
+# Vocabulary Types
+
+We refer to standard library types like `std::array`, `std::map`, `std::optional`,  `std::variant` and `std::vector` as _vocabulary types_.
```
nbx8, line 43: We refer to standard library types such as

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+We'd anticipate that an arbitrary piece of C++ library or application code would make use of some of these types.
```
nbx8, line 45: Maybe this?

We postulate that an arbitrary piece of C++ library or application code would make use of some of these types.

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+Standardizing vocabulary types is important as it allows different libraries to easily interoperate and for users to build applications.
+
+The standard library contains other, non-vocabulary types to do specific jobs like interacting with the filesystem, formatting text for output, or dealing with concurrency.
```
nbx8, line 49: ... to do specific jobs such as interacting...

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+We probably want to standardize both sorts of library type.
+
+---
+
+# Composite classes
+
+We can define composite classes: classes with other class instances as member data.
```
nbx8, line 57: We define composite classes as: classes with other class instances as member data.

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+# Incomplete types
+
+We may want a composite to contain an instance of an incomplete type, either directly or through the use of existing vocabulary types.
+
+Incomplete type support is needed for defining recursive data structures, supporting open-set polymorphism, and hiding implementation detail.
+
+Incomplete types are supported by storing the object of the incomplete type outside of the composite object.
+
+Storing an object outside of a composite type can also be used to optimize use of cache (hot-cold-splitting).
```
nbx8, line 136: hot/cold splitting 
hot cold splitting
hot-cold splitting

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+pointers are the first thing we may reach for:
+
+```c++
+class Composite {
+    T* ptr_;
+    // ...
+}
+```
+
+Pointers are a poor fit for owned data, we'll need to implement all special member functions and manually check const-qualified methods - const does not propagate to a pointee when the pointer is accessed through a const-access path.
```
nbx8, line 152: Pointers are a poor fit for owned data as we need to implement all special member functions and manually check const-qualified methods (const does not propagate to a pointee when the pointer is accessed through a const-access path).

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+Unique pointers are a little better:
+
+```c++
+class Composite {
+    std::unique_ptr<T> ptr_;
+    // ...
+}
+```
+
+We don't need to implement move construction, move assignment or destruction. The compiler will implicitly delete the copy constructor and copy assignment operator; either the composite is non-copyable or we implement the copy constructor and copy assignment operator ourselves.
```
nbx8, line 167: We do not need to implement ...

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+Const does not propagate through `std::unique_ptr`, we must manually check const-qualified methods for correctness.
+
+---
+
+## Incomplete types using `std::shared_ptr<T>`
+
+Shared pointers don't model the right thing:
```
nbx8, line 175: Shared pointers do not model the right thing:

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+```c++
+class Composite {
+    std::shared_ptr<T> ptr_;
+    // ...
+}
+```
+
+We don't need to implement any special member functions, but we have other issues.
```
nbx8, line 184: We do not need to implement...

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+
+The compiler-generated copy constructor and assignment operator will give rise to multiple composite objects that share the same components: that's not ownership.
```
nbx8, line 186: this is not ownership

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+Shared pointers to const might work:
+
+```c++
+class Composite {
+    std::shared_ptr<const T> ptr_;
+    // ...
+}
+```
+
+Again the compiler-generated copy constructor and assignment operator will give rise to multiple composite objects that share the same components, _but_ the shared components
```
nbx8, line 203: Again, ...

presentations/vocabulary_types_2023_11_8.md
```diff
@@ -0,0 +1,241 @@
...
+    // ...
+}
+```
+
+Again the compiler-generated copy constructor and assignment operator will give rise to multiple composite objects that share the same components, _but_ the shared components 
+are immutable. 
+
+If our hierarchy of composites is const-correct, immutability will hide the fact that data is shared.
+
+We can't call non-const qualified member function on our component accessed though `ptr_`.
```
nbx8, line 208: We cannot call ...

nbx8 (dismissed):
Clear, informative slides. Only minor edits suggested.

presentations/vocabulary_types_2023_11_8.md
jbcoe: I went with "We define composite classes as classes with other class instances as member data."

Twon (approved):

PR #122, opened 2023-11-07
Move benchmarks and compile_checks into subdirectories

jbcoe:
Add missing benchmarks that were overlooked in the large BUILD file.


Twon (approved):
Looks like the compile_checks are only run by Bazel.  We might consider adding these to CMake in the future.

But looks like a sensible change.

PR #130, opened 2023-11-08
Dev twon clang format

Twon:


Twon:
Addressed in https://github.com/jbcoe/value_types/pull/129

PR #134, opened 2023-11-08
Add compile_checks to CMakeLists.txt

jbcoe:
Exclude compile checks from MSVC as there are issues with constexpr/consteval


Twon (dismissed):
Nice, it will be good to support these via CMake

jbcoe:
@Twon weird Visual Studio build errors that I can't decipher - can you repro locally?

Twon:
> @Twon weird Visual Studio build errors that I can't decipher - can you repro locally?

Looking into it, can reproduce locally

jbcoe:
> > @Twon weird Visual Studio build errors that I can't decipher - can you repro locally?
> 
> Looking into it, can reproduce locally

There have been bugs: https://developercommunity.visualstudio.com/t/problem-with-consteval-function/1224555

Twon:
> > @Twon weird Visual Studio build errors that I can't decipher - can you repro locally?
> 
> Looking into it, can reproduce locally

Looks like a compiler bug... I've cut this down to a minimal example and it still fails with no error message.  Will try to create an example on Godbolt

Twon:
> > > @Twon weird Visual Studio build errors that I can't decipher - can you repro locally?
> > 
> > 
> > Looking into it, can reproduce locally
> 
> Looks like a compiler bug... I've cut this down to a minimal example and it still fails with no error message. Will try to create an example on Godbolt

Reproduction case on Godbolt: https://godbolt.org/z/Ko3EMoo1K

jbcoe:
@Twon Can we exclude these from MSVC for now? My CMake-fu is weak

Twon (approved):

PR #152, opened 2023-11-09
Discuss design of allocator aware types

jbcoe:


DRAFT.md
```diff
@@ -216,15 +216,20 @@ Note: As the null state of `indirect` and `polymorphic` is not observable, and
...
 by implementers to exchange pointers on move construction and assignment.
 
-### Noexcept
-
-TODO
-
-### Allocators as template arguments
+### Allocator support
+
+Both `indirect` and `polymorphic` are allocator-aware types. They must suitable
```
villevoutilainen, line 221: They must <ins>be</ins> suitable

DRAFT.md
```diff
@@ -216,15 +216,20 @@ Note: As the null state of `indirect` and `polymorphic` is not observable, and
...
+for use in allocator-aware composite types and containers. Existing
+allocator-aware types in the standard, such as `vector`, `map` and `tuple`, take
```
villevoutilainen, line 223: std::tuple doesn't have a template parameter for an allocator, because it doesn't store it.

But indirect could be seen to be similar to the proposed std::basic_optional. If you make an indirect valueless, and then put a value into it, perhaps it needs to use a remembered allocator to construct the new payload?

villevoutilainen (changes requested):

DRAFT.md
jbcoe: If you make an indirect valueless, and then put a value into it, it will use a remembered allocator to construct the new payload. How do I write wording to say this? Is it not implicit?

nbx8 (approved):
Clearly stated design motivation.

DRAFT.md
villevoutilainen: You would say that by changing the specification of the assignment operator to say that it performs allocator-construction with the stored allocator, instead of
"Otherwise if *this is valueless, constructs an owned object using the copy constructor of the object owned by other." So that bit looks incorrect to boot. Saying "using the copy constructor" is bogus, you should say what expression is used, and that just might not resolve to a copy constructor. You'll need to say it uses  uninitialized_construct_using_allocator().

jbcoe: Might you post proposed wording as an edit as part of this PR?

villevoutilainen: You need a Nina Ranns on this. She has worked on implementing and specifying an allocator-aware optional and variant, she can point you to all of the problems.

jbcoe: I've updated working. It might benefit from LWG review but is improved for now.

PR #153, opened 2023-11-09
Discuss operator() operator[] and emplace

### LINKED ISSUE #143: Discuss `emplace` method

### LINKED ISSUE #145: Add discussion of `operator[]` and `operator()`

jbcoe:
Fixes #143
Fixes #145

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -228,13 +228,52 @@ TODO
 
 TODO
 
-### Extra constructors and `emplace`
+### Supporting `operator()` `operator[]`
 
-TODO
+Neither `indirect` nor `polymorphic` support `operator()` or `operator[]`. These
+operators can't be implemented with `=default` and cannot, in general, be
+meaningfully implemented for a composite by aggregating over its components.
+These functions are not core to the design of either class template; both could
+be added as non-breaking changes in a later revision of the C++ standard if
+required.
 
-### Forwarding functions for indirect
+### Member function `emplace`
 
-TODO
+Neither `indirect` nor `emplace` support `emplace` as a member function.
+The member function `emplace` could be added as :
+
+```c++
+template <typename ...Ts>
+indirect::emplace(Ts&& ...ts);
+```
+
+```c++
+template <typename U, typename ...Ts>
+polymorphic::emplace(in_place_type<U>, Ts&& ...ts);
+```
+
+This would be API noise. It offers no efficiency improvement over:
+
+```c++
+some_indirect = indirect(/* arguments */);
+```
+
+```c++
+some_polymorphic = polymorphic(in_place_type<U>, /* arguments */);
+```
+
+If `emplace` was desirable to avoid repeating the type name, it might be better
+implemented as a non-member function:
+
+```c++
+template <typename T, typename ...Ts>
+void emplace(T& t, Ts&& ...ts) {
+  t = T(std::forward<Ts>... ts);
+}
+```
+
+We don't propose adding a `emplace` as a non-member function to the C++ standard
+libary as part of this proposal.
 
 ### Design for polymorphic types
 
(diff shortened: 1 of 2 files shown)


DRAFT.md
```diff
@@ -228,13 +228,52 @@ TODO
 
 TODO
 
-### Extra constructors and `emplace`
+### Supporting `operator()` `operator[]`
 
-TODO
+Neither `indirect` nor `polymorphic` support `operator()` or `operator[]`. These
+operators can't be implemented with `=default` and cannot, in general, be
+meaningfully implemented for a composite by aggregating over its components.
```
villevoutilainen, line 235: I don't think this is a good explanation. an operator() can be made available if the (single) element of indirect has it, but the reason for not providing it is that it's always possible to just get the value and apply the call operator. That rationale doesn't apply to, say, operator<=>, because valueless indirects presumably order among the others? So you can't just <=> the value, you have to <=> the indirect. But for something like operator(), you can just always "dereference" the indirect.

villevoutilainen: Oh, you can't spaceship valueless indirects either. Oh well.

DRAFT.md
```diff
@@ -228,13 +228,52 @@ TODO
 
 TODO
 
-### Extra constructors and `emplace`
+### Supporting `operator()` `operator[]`
 
-TODO
+Neither `indirect` nor `polymorphic` support `operator()` or `operator[]`. These
+operators can't be implemented with `=default` and cannot, in general, be
```
nbx8, line 234: These operators cannot be implemented with ...

DRAFT.md
```diff
@@ -228,13 +228,52 @@ TODO
...
+meaningfully implemented for a composite by aggregating over its components.
+These functions are not core to the design of either class template; both could
+be added as non-breaking changes in a later revision of the C++ standard if
+required.
 
-### Forwarding functions for indirect
+### Member function `emplace`
 
-TODO
+Neither `indirect` nor `emplace` support `emplace` as a member function.
```
nbx8, line 242: Neither  `indirect` nor `polymorphic` support `emplace` as a member function.

DRAFT.md
```diff
@@ -228,13 +228,52 @@ TODO
...
+implemented as a non-member function:
+
+```c++
+template <typename T, typename ...Ts>
+void emplace(T& t, Ts&& ...ts) {
+  t = T(std::forward<Ts>... ts);
+}
+```
+
+We don't propose adding a `emplace` as a non-member function to the C++ standard
```
nbx8, line 275: We do not...

nbx8 (approved):

DRAFT.md
jbcoe: Nicely spotted!

DRAFT.md
jbcoe: Is there a neater argument? Keen to merge unless what I've said is flat-out wrong.

villevoutilainen: Well, it's plain incorrect to say that such an operator couldn't be done in terms of the component, because there's only one.

jbcoe: I mean :

```c++
class Composite {
  indirect<A> a_;
  indirect<B> b_;
  
  auto operator() {
     // ???
   }
 };
 ```
 
 How can I reword my text to reflect this?

Twon (approved):

DRAFT.md
villevoutilainen: Ahh.. yes. Something like this:
"There's no need for indirect/polymorphic to provide a function call or an indexing operator. Users who wish to do that can just access the value and call its operator. Furthermore, unlike comparisons, function calls or indexing operators don't compose further; for example <that example of yours>, a Composite wouldn't be able to automatically generate a composited operator() or an operator[]."

jbcoe: Very nice. Applied.

PR #155, opened 2023-11-10
Dev jbcoe kona allocator wording

jbcoe:
@NinaRanns does this add (some of) the right wording in the right place for allocators? We will still to add/update _Effects_ wording for various functions but I can do that in a later PR.


jbcoe:
Discussed offline with @villevoutilainen. Merging.

jbcoe:
Discussed offline with @villevoutilainen. Merging.

PR #160, opened 2023-11-10
Fix swap and assign for allocator propagation traits

jbcoe:
I think that we've got swap and assignment working correctly now. We may need more tests to increase coverage, I'll add these in a follow-up PR.

Update: Coverage pushed to 100%.


Twon (approved):

jbcoe:
Discussed offline with @villevoutilainen. Merging.

PR #173, opened 2023-11-10
Run pre-commit checks as GitHub Action workflows

jbcoe:


Twon (changes requested):
Please reinstate the integrity checks. This contains the action validator which requires rust dependencies installed locally to run.  We run these on CI as there is a github action which installs these dependencies in the docker environment of the github workflow.

Twon (approved):

PR #194, opened 2023-11-12
Prepare allocator aware control block for polymorphic

jbcoe:


jbcoe:
rendered obsolete by #249

PR #195, opened 2023-11-12
Check allocator create/destroy for indirect to test swap and assign

jbcoe:


indirect_test.cc
```diff
@@ -614,4 +614,173 @@ TEST(IndirectTest, FormatNativeTypesPropagateFormatting) {
...
+  }
+
+  void deallocate(T* p, std::size_t n) {
+    std::allocator<T> default_allocator{};
+    default_allocator.deallocate(p, n);
+  }
+
+  template <typename... Ts>
+  void construct(T* p, Ts&&... ts) {
+    new (p) T(std::forward<Ts>(ts)...);
```
Twon, line 678: This could be [`std::construct_at`](https://en.cppreference.com/w/cpp/memory/construct_at), then this could be used in constexpr tests

indirect_test.cc
```diff
@@ -614,4 +614,173 @@ TEST(IndirectTest, FormatNativeTypesPropagateFormatting) {
...
+  }
+
+  void destroy(T* p) {
+    if constexpr (Taggable<T>) {
+      if (p->tag() != tag_) {
+        throw TagMismatch(tag_, p->tag());
+      }
+    } else /* constexpr */ {
+    }
+    p->~T();
```
Twon, line 692: Similarly could be [`std::destory_at`](https://en.cppreference.com/w/cpp/memory/destroy_at)

Twon (approved):
Minor suggestions, otherwise looks good

indirect_test.cc
jbcoe: I can't get this to work. Will merge now as it's non-blocking. We can tackle if needed in another PR.

PR #196, opened 2023-11-12
Define `unreachable` only once, even if including both headers

Ukilele:
Right now, including both indirect.h and polymorphic.h will lead to a compiler error: `error: redefinition of 'unreachable'`. Because both headers define the function `unreachable`. By wrapping it in a macro, we ensure that it will only be defined once.


jbcoe:
It might be better still to move unreachable to its own file. This PR improves things regardless.

jbcoe (approved):

Twon (approved):

PR #197, opened 2023-11-12
Minor fixes/suggestions

lewissbaker:
Reviewed up to indirect.ctor.

- Some changes of free-store-allocated to dynamically-allocated in abstract/design sections.
  I think that these types permit small-object optimisations, in which case they may not be on the free-store?
  Also, an allocator could allocate them in stack storage, so may not necessarily be on the “free store”.
- Simplified paragraph in abstract relating to const-propagation.
   This type is not necessarily only used as a member of a parent object.
   Just talk about constness of `polymorphic<T>` propagating the constness to the `T` rather than talking about constness of the parent object influencing that.
- Remove sentence about allocator requiring that move construction is non-throwing.
  I think this is already guaranteed by the requirements of Cpp17Allocator and so may not need to be stated here.
- Added extra paragraph to section about default constructibility clarifying that the indirect/polymorphic types are `std::is_default_constructible_v<T>` but that attempting to odr-use the default-constructor would be I’ll-formed rather than just saying “is default constructible”.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -19,7 +19,7 @@ _Sean Parent \<<sparent@adobe.com>\>_
 We propose the addition of two new class templates to the C++ Standard Library:
 `indirect<T>` and `polymorphic<T>`.
 
-These class templates have value semantics and compose well with other standard
+Specializations of these class templates have value semantics and compose well with other standard
 library types (such as vector) allowing the compiler to correctly generate
 special member functions.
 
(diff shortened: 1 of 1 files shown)


DRAFT.md
```diff
@@ -160,6 +158,12 @@ logical state nor to the logical state of other object.
 is default constructible. Moving a value type onto the free store should not add
 or remove the ability to be default constructed.
 
+Note that, due to the requirement to support incomplete `T` types, the `indirect<T>`
```
lewissbaker, line 161: Did you explore defining the default-constructor using something like:
```
template<typename U = T>
indirect() requires std::constructible_from<U>;
```

this would allow defering the computation of whether the T is default-constructible until it’s actually used/referenced.

e.g. https://godbolt.org/z/q6dEn3Wr9

You get into trouble if anyone actually asks that question while T is incomplete.
i.e. asks `std::constructible_from<indirect<T>>` while T is incomplete (it will report false, even if T later turns out to have a default-constructor once defined).
If you uncomment the static-assert in the godbolt example above then it makes the program ill-formed - not because the static_assert fails, but because it forces the compiler to compute the fact that the default constructor is not present, then later when you define the T and try to use indirect<T> default constructor, there is none defined.

Attempts to odr-use the default constructor while the T is incomplete will always be a compile-error, however.

jbcoe (approved):

DRAFT.md
lewissbaker: @jbcoe just checking - did you see this comment (you merged a few minutes after I wrote it)
do you have any thoughts on this alternative approach?

jbcoe:
I'll look into it, I'll raise an issue to remember

thanks

Jon

DRAFT.md
jbcoe: I added https://github.com/jbcoe/value_types/issues/199

PR #200, opened 2023-11-12
Constraints with incomplete types

jbcoe:


indirect.h
```diff
@@ -66,15 +66,18 @@ class indirect {
   using pointer = typename allocator_traits::pointer;
   using const_pointer = typename allocator_traits::const_pointer;
 
+  template <typename U = T>
   constexpr indirect()
-    requires std::is_default_constructible_v<A>
+    requires(std::is_default_constructible_v<A> &&
+             std::is_default_constructible_v<U>)
```
Twon, line 72: This delays the evaluation of the constraint until this function is instantiated so it works with incomplete types?

indirect.h
```diff
@@ -66,15 +66,18 @@ class indirect {
...
   {
-    static_assert(std::is_default_constructible_v<T>);
     p_ = construct_from(alloc_);
   }
 
-  constexpr indirect(std::allocator_arg_t, const A& alloc) : alloc_(alloc) {
-    static_assert(std::is_default_constructible_v<T>);
+  template <typename U = T>
+  constexpr indirect(std::allocator_arg_t, const A& alloc)
+    requires std::is_default_constructible_v<U>
```
Twon, line 79: Perhaps we should be checking that the allocator is copy-constructible here too?

Twon (approved):
A couple of minor points/queries. otherwise looks good.

indirect.h
jbcoe: That’s my understanding. I’m not sure if that means we can graduate our _Mandates_ to a _Constraints_

indirect.h
jbcoe: From https://en.cppreference.com/w/cpp/named_req/Allocator this is a requirement of any allocator. It just needs to satisfy the _Cpp17AllocatorRequirments_.

Nice catch though.

jbcoe: Do we state allocator requirements anywhere?

indirect.h
jbcoe: @lewissbaker does this PR mean we can/should change wording too?

lewissbaker: Yes. I think the wording would need to be updated to use Constraints.
The signature would probably also need to be updated to:
```
template<typename U = T>
constexpr indirect();
```
Constraints: `is_default_constructible_v<U>` is `true` and `is_default_constructible_v<Alloc>` is `true`.

indirect.h
lewissbaker: The front matter has a requirement that Alloc satisfy Cpp17Allocator concept, I believe.

indirect.h
```diff
@@ -66,15 +66,18 @@ class indirect {
   using pointer = typename allocator_traits::pointer;
   using const_pointer = typename allocator_traits::const_pointer;
 
+  template <typename U = T>
   constexpr indirect()
```
lewissbaker, line 70: One question here is whether the default-constructor of `indirect()` should be doing _Uses-allocator construction_ of the object. i.e. if `T` supports a  `std::allocator_arg_t` constructor then it calls that with the default-constructed allocator instead of calling the default constructor.

indirect.h
lewissbaker: > The front matter has a requirement that Alloc satisfy Cpp17Allocator concept, I believe.

Hmm, no it seems it's not there. I have added #227 to move this requirement to the front-matter.

lewissbaker:
Is it also worth applying the same approach to the other special-member functions?

e.g. to conditionally define copy-construction / copy-assignment / move-assignment if the object is copy-constructible/move-constructible?

jbcoe:
This approach works and should be applied everywhere.

jbcoe:
I can't apply this approach to `operator = ` or to copy/move constructors without running into compiler errors.

I'm weakly inclined to say that we should use mandates consistently across special member functions. 

@lewissbaker what do you think?

lewissbaker:
Yeah, I haven't been able to figure out how to make constraints work for the other special member functions, so maybe it's best just to go with mandates for now.

I wonder what the implications are for other types that contain an `indirect` data-member and that end up with defaulted special-member functions? e.g. say I have `struct foo { indirect<move_only> data; };`, this type would gain a compiler-generated defaulted copy-constructor/copy-assignment, even though odr-using that would be ill-formed.

jbcoe:
This change has been implemented and merged in another PR.

PR #204, opened 2023-11-13
Use xyz::guard in indirect and polymorphic

jbcoe:
@Twon Do we want this or does it raise cognitive load for readers of what is mean to be a reference implementation?

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -25,6 +25,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <compare>
 #include <concepts>
 #include <memory>
+#include <optional>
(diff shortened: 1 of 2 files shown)


Twon (approved):

Twon:
> @Twon Do we want this or does it raise cognitive load for readers of what is mean to be a reference implementation?

Looks sensible from an exception safety point of view!

jbcoe:
I'll press on then. Will finish this tonight. Code was exception-safe before.

Twon:
> Code was exception-safe before.

Of course it was... lack of sleep due to jet lag is getting to me :(

jbcoe:
Using guard in C++14 implementation looks a bit trickier, some issues with RVO. Will defer for now.

PR #206, opened 2023-11-13
Add death test for non-propagating swap when allocators are not equal

### LINKED ISSUE #202: swapping with unequal allocators calls unreachable
Following code shows a test that erroneously fails.
```c++
TEST(Indirect, AllocatorSwapIssue) {
  auto* resourceA = std::pmr::new_delete_resource();
  unsigned char buffer[1024];
  std::pmr::monotonic_buffer_resource resourceB(buffer, 1024);

  std::pmr::polymorphic_allocator allocA(resourceA);
  std::pmr::polymorphic_allocator allocB(&resourceB);

  xyz::indirect<int, std::pmr::polymorphic_allocator<int>> iA(
      std::allocator_arg, allocA, 10);
  xyz::indirect<int, std::pmr::polymorphic_allocator<int>> iB(
      std::allocator_arg, allocB, 220);

  iA.swap(iB);
  EXPECT_EQ(*iA, 220);
  EXPECT_EQ(*iB, 10);
}
```
We have two indirects `iA` and `iB` using a `std::pmr::polymorphic_allocator<int>` where both allocators are constructed from a different resource, such that `alloc_ == other.alloc_` returns `false`. When calling `iA.swap(iB)` then we currently hit the call to `unreachable` at https://github.com/jbcoe/value_types/blob/main/indirect.h#L224 . The same issue applies to `polymorphic`.

jbcoe:
Adds a test for UB in swap with on-equal allocators

Fixes #202

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -120,4 +120,4 @@ jobs:
         working-directory: ${{github.workspace}}/build
         # Execute tests defined by the CMake configuration.
         # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
-        run: ctest -C ${{ matrix.configuration }}
+        run: ctest -C ${{ matrix.configuration }} --output-on-failure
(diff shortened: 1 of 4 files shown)


Twon (approved):

jbcoe:
I'm not sinking more time into getting this working. Will close.

PR #208, opened 2023-11-13
Add unprofiled hidden-inline macro for indirect and polymorphic

jbcoe:
We need proof that this works before we consider merging it.

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -34,9 +34,18 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 namespace xyz {
 
+#ifndef XYZ_HIDDEN_INLINED
+#if defined(_MSC_VER)
+#define XYZ_HIDDEN_INLINED
+#else
+#define XYZ_HIDDEN_INLINED \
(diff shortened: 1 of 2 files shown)


Twon:
Why not just set the [VISIBILITY_INLINES_HIDDEN](https://cmake.org/cmake/help/latest/prop_tgt/VISIBILITY_INLINES_HIDDEN.html#prop_tgt:VISIBILITY_INLINES_HIDDEN) target property for CMake.  This will set the `-fvisibility-inlines-hidden` flag for the target, which pesumably there is an equivalent mechanism in Bazel?

Twon:
Btw, what is the motivation for this change?  My understanding is this will ensure these symbols are hidden within a shared object [C++ -fvisibility=hidden -fvisibility-inlines-hidden](https://stackoverflow.com/questions/3570355/c-fvisibility-hidden-fvisibility-inlines-hidden).  But as we are not using shared objects, are we trying to propagate this onto users to do build these types into a shared object?

jbcoe:
This change was motivated by curiosity and inability to sleep on a flight. ✈️ 

Will close.

PR #209, opened 2023-11-14
Fix constraints wording for comparisons

jbcoe:
Optional has: 

"Mandates: The expression *x == *y is well-formed and its result is convertible to bool. [Note: T need
not be Cpp17EqualityComparable. — end note]"

https://isocpp.org/files/papers/N4860.pdf


Twon (approved):
Odd this PR shows as having no files in it?

But the changes to `draft.md` in the commits look good

PR #215, opened 2023-11-14
Add C++11 implementations of polymorphic and indirect

jbcoe:
One very tricky aspect will be that the tests themselves need C++17 (I think). We need a cunning way to check that C++11 code really compiles with a C++11 compiler. Given that the indirect and polymorphic classes are templates this might mean writing extra tests for them (just including the header in an empty c++11 project might not be enough); I don't have a solution yet.


jbcoe:
@Ukilele I pushed some Bazel changes and tweaks to your headers. We have two failing build targets now.
I'm not sure if you use Bazel or CMake but you can make similar changes to CMakeLists.txt to get targets added there too.

Let me know if anything is unclear.

Ukilele:
I think the reason for the failures is, that I did not add the type `xyz::NoPolymorphicSBO`, right? I am not sure what this type is about. It does get defined in the [polymorphic.h](https://github.com/jbcoe/value_types/blob/dev-ukilele-c%2B%2B11/polymorphic.h#L56C8-L56C25). But it seems like it is a type used for unit tests only. So shouldn't it be moved to the unit-test files instead?

Ukilele:
I've added a new header `compatibility/utility.h`. As C++11 does not provide a `std::in_place_type(_t)`, I added `xyz::in_place_type_t` to `utility.h`. It falls back to `std::in_place_type_t` in case we have it. This header needs to be included by `polymorphic_cxx11.h` which works fine. But then I went to update the unit tests in `polymorphic_test.cc` to use `xyz::in_place_type_t<...>{}` instead of `std::in_place_type<...>` to work in both C++20 and C++11 mode. But now the `polymorphic_test.cc` has to always (also in C++20) include `utility.h` to get the definition of `xyz::in_place_type_t`. But I can't get this working with Bazel. Help appreciated :)

jbcoe:
I pushed some fixes and renames. This is the right way to work: push what you have and get help early.

Ukilele:
Thanks a lot for your help Jon! :)
From my side everything looks fine and ready to merge now. But please give it a thorough review first ;)

jbcoe:
I think we might want to keep the headers isolated so that they can be included in compiler explorer as a single file. I'll work on this - will involve some duplication.

jbcoe:
These look great to me. Ideally we'd have some C++11-only tests to exercise them along with test coverage. I'm sure folk will raise issues if there are incompatibilities. Great work turning this round so fast.

jbcoe:
Nicely spotted!

Ukilele:
> One very tricky aspect will be that the tests themselves need C++17 (I think). We need a cunning way to check that C++11 code really compiles with a C++11 compiler. Given that the indirect and polymorphic classes are templates this might mean writing extra tests for them (just including the header in an empty c++11 project might not be enough); I don't have a solution yet.

@jbcoe , we could check which test cases require C++>11 (e.g. the ones using std::optional) and wrap them in an `#ifdef` such that they get only compiled and executed when we have an appropriate compiler. And then we could add some additional CI-tests that use some old C++11 compilers, define `XYZ_POLYMORPHIC_CXX_11` and `XYZ_INDIRECT_CXX_11` such that the c++11-implementations get used and run the tests in C++11 mode. By this approach we would have one test file (for polymorphic and indirect respectively) (with several ifdefs), which I think is desirable. What do you think?

jbcoe:
One test file with ifdefs feels lke the right way to go. 

GoogleTest itself might need C++14 as a minimum, maybe we should bump the indirect/polymorphic headers to use C++14 otherwise I'm not sure the tests themselves will build.

Adding `copts` to `bazel` targets should be enough to force a C++ version: `-std=c++14`.

Let me know if you need further input.

jbcoe:
Bazel tests reveal a handful of fixes but this direction looks good. I can add appropriate targets to CMake this evening and push a PR.

Twon:
Looking into the CMake now, will push up as soon as I have something building

Ukilele:
Ok the tests should run with C++14 now, regardless of if they use `polymorphic.h` or `polymorphic_cxx11.h` (same for indirect).
gtest indeed seems to need C++14. I don't know which version of gtest we use, but version 1.10.0 is fine with C++11 (at least I am almost sure about it).
I also use `std::is_final` in the Empty-Base-Optimization which requires C++14. I could use `__is_final` for GCC and Clang and just not use the feature at all under MSVC (in case we even have to support such old MSVC compilers). But I guess for now I will leave things as is.

jbcoe:
I made a handful of changes so that headers can work as single-file-includes in https://godbolt.org. 

Most folk will use THESE headers as C++20 is still relatively new. Great work @Ukilele .

jbcoe:
> I think the reason for the failures is, that I did not add the type `xyz::NoPolymorphicSBO`, right? I am not sure what this type is about. It does get defined in the [polymorphic.h](https://github.com/jbcoe/value_types/blob/dev-ukilele-c%2B%2B11/polymorphic.h#L56C8-L56C25). But it seems like it is a type used for unit tests only. So shouldn't it be moved to the unit-test files instead?

The existence of this class is a simple way of the SBO implementation turning off SBO for some types. It needs to be a complete type in the SBO impl and in tests, I don't see a simple way of isolating the class definition.

Twon (approved):

Twon:
Fantastic work @Ukilele!  Thanks for your efforts!

PR #217, opened 2023-11-15
Use 'shall' for requirements placed on the user

lewissbaker:
Requirements on the user should use the word 'shall' rather than 'must'.


Twon (approved):

jbcoe (approved):
Thanks for this.

PR #223, opened 2023-11-15
Simplify some constructors by delegating to other constructors

lewissbaker:
For the copy/alloc+copy/alloc+move constructors specify the effects in terms of other constructors as a way of simplifying the specification.

NOTE: That this PR also makes the `indirect(std::allocator_arg_t, const Allocator& alloc, indirect&& other)` now conditionally `noexcept`. This is required because the case where `alloc != other.get_allocator()` may require allocating a new object and move-constructing the new object from `*other`.

The main question here is whether the post-condition that `other` is valueless is required to hold.
If it allocates a new object rather than taking ownership of `other`'s object, do we want to force `other` to become valueless, or just leave it having a value that is in a moved-from state?

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -600,12 +600,10 @@ constexpr indirect(const indirect& other);
 
 * _Preconditions_: `other` is not valueless.
 
-* _Effects_: Constructs an indirect owning an instance of `T` created with the
-  copy constructor of the object owned by `other`. `allocator` is obtained by
-  calling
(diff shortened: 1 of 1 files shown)


jbcoe (approved):

jbcoe (approved):

jbcoe:
I don't think we should force a moved from object to be valueless if allocator mismatch means it can be just moved from.
It's extra work for no benefit. The standard gives us few guarantees about the state of the moves from object but we know it's non-null so valueless_after_move should return false.

PR #224, opened 2023-11-15
Fix signatures of `indirect` comparison operators

lewissbaker:
Use the `Allocator` template parameter instead of the undefined `A` name.


Twon (approved):

jbcoe (approved):
Thanks for this.

PR #225, opened 2023-11-15
Tweak wording for `indirect` comparison operators

lewissbaker:
The Constraints section of the wording was trying to say that two types were comparable, rather than two const-ref lvalues to objects of those types being comparable.


Twon (approved):

jbcoe (approved):
Thanks for this.

PR #227, opened 2023-11-15
Move requirement for Allocator to meet Cpp17Allocator requirements to front matter

lewissbaker:
Add this requirement to the front-matter of `indirect` and `polymorphic` and remove it from the preconditions of each of the various member functions that might allocate.


jbcoe (approved):
Thanks for this.

PR #237, opened 2023-11-15
Add code-font where necessary and other minor typos in wording

lewissbaker:
- add code-font to `true` and `false`
- Change `allocator` -> `allocator_` where it is referring to the exposition-only member.
- Parenthesize `(*this).get_allocator()`


jbcoe (approved):
Great work Lewis, thanks so much for this!

PR #244, opened 2023-11-17
Remove SBO-specific detail from polymorphic

jbcoe:
Allow CMake tests to be made manual-only (SBO tests fail).

### DIFF
--- a/cmake/vt_add_test.cmake
+++ b/cmake/vt_add_test.cmake
@@ -78,9 +78,11 @@ function(vt_add_test)
     endif()
 
     include(GoogleTest)
-    gtest_discover_tests(${VALUE_TYPES_NAME})
+    if(NOT ${VALUE_TYPES_MANUAL})
+        gtest_discover_tests(${VALUE_TYPES_NAME})
+    endif()
 
-    if (ENABLE_CODE_COVERAGE)
+    if (ENABLE_CODE_COVERAGE AND NOT ${VALUE_TYPES_NAME})
         add_coverage(${VALUE_TYPES_NAME})
     endif()
 
--- a/cmake/vt_add_test.cmake
+++ b/cmake/vt_add_test.cmake
@@ -23,7 +23,7 @@ associates patterns and allows configuration for common optional settings
 #]=======================================================================]
 function(vt_add_test)
     set(options)
-    set(oneValueArgs NAME)
+    set(oneValueArgs NAME MANUAL)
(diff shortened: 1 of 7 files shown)


jbcoe:
@Ukilele I think this clears up some of the SBO-mess I made on the plane.

cmake/vt_add_test.cmake
```diff
@@ -23,7 +23,7 @@ associates patterns and allows configuration for common optional settings
 #]=======================================================================]
 function(vt_add_test)
     set(options)
-    set(oneValueArgs NAME)
+    set(oneValueArgs NAME MANUAL)
```
Twon, line 26: If you are just setting this to true then this would make more sense to be an `option` than a `oneValueArgs` as you only care about its presence, not the actual value.

cmake/vt_add_test.cmake
```diff
@@ -23,7 +23,7 @@ associates patterns and allows configuration for common optional settings
 #]=======================================================================]
 function(vt_add_test)
     set(options)
```
Twon, line 25:
```suggestion
    set(options MANUAL)
```

cmake/vt_add_test.cmake
Twon, line 26:
```suggestion
    set(oneValueArgs NAME)
```

cmake/vt_add_test.cmake
```diff
@@ -64,11 +64,15 @@ function(vt_add_test)
...
+    else()
+        add_test(
+            NAME ${VALUE_TYPES_NAME}
+            COMMAND ${VALUE_TYPES_NAME}
+            WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+        )
+    endif()
 
     include(GoogleTest)
     gtest_discover_tests(${VALUE_TYPES_NAME})
```
Twon, line 78: I don't this it make sense to call `gtest_discover_tests` it you have not called the branch that calls `add_test`

cmake/vt_add_test.cmake
Twon: Also can you document the `MANUAL` parameter in the function documentation

jbcoe: Done.

cmake/vt_add_test.cmake
jbcoe: Fixed.

Twon (approved):

PR #246, opened 2023-11-17
Ben Craig's quick review of wording

ben-craig:
I made an editing pass on the wording.  There's likely aspects that I did wrong, so pick and choose what you keep accordingly.

Most of the things I saw where LWG and editorial level, but you might as well fix those while you are in the period between meetings.

The biggest design'ish thing I saw that needs to be resolved is constraints vs. mandates.  I've got a note about that below.  The biggest wording thing I saw that I didn't attempt to fix was the possible technique of declaring these facilities as allocator-aware containers.  That has the potential to remove lots of words.

Below are my stream of consciousness review notes.

--------------------------------------------
typesetting with .md may be a challenge.  Maybe the typesetting requirements for this wording aren't that bad though.

Added paragraph numbers.  Editors can change them, but they make referring to specific paragraphs during reviews much easier.  For example...

I think indirect.general#2 and indirect.general#3 can be replaced by saying "The class template indirect satisfies the requirements of an allocator-aware container [container.alloc.reqmts]."  basic_stacktrace and regex match_results would be precedent, as they are non-container allocator-aware container.  regex wording is old though, so don't look to that for much precedent.

indirect.general#5 why the ban on unions?  There's no other mention of unions in the paper.

Cleaned up some uses of "shall".  "shall" is used to describe requirements that the user must satisfy.  So saying that some user supplied type shall meet some named requirement is fine.  When you need the implementation to do something, you express that by stating a fact, and saying what "is".  "Otherwise, the allocators shall not be swapped" suggests the user has direct control, so that was replaced with "Otherwise, the allocators are not swapped".

Do you need reference and const_reference typedefs?  I think the answer is no.  Those typedefs are useful on "real" containers due to container adaptors, and maybe because of iterators?  But "reference" isn't something that's usefully grepable in the standard right now.

You have an exposition-only `T* p_;`.  You don't use this consistently.  For example, the default indirect ctor doesn't set it.  You rely on it being set in the effects of [indirect.assign] though.  So you either need to use it consistently, or remove it.

Variadic construction: https://eel.is/c++draft/utilities#variant.mod-7
I suspect you want to forward arguments.  You also need to specify list or non-list initialization (likely non-list).  Look for the term direct-non-list-initialized.

Do you have thorough testing of indirect<indirect<T>>?  I'm worried that it may do nasty things to the ctor overload set.

I wonder if you can get rid of most of [indirect.assign] and [indirect.swap], based on being an allocator-aware container.  basic_stacktrace's op= is pretty sparse.

Is there rationale for your choice of "constrains" vs. "mandates" in your ctors?  Which type are you modeling for your choice of SFINAE vs. static_asserts?  variant and optional seem to prefer constrains, but optional copy has a mandate-like thing in remarks.  I have a very weak preference for mandates, as I think they provide better diagnostics at the expense of meta-programming opportunities, but I'll go with whichever approach has rationale.

I'm not super familiar with std::format extension.  The "format" extension point has precedent with chrono, but the stuff chrono does for parse is different than what you have.  No idea if that's a problem or not.

Design warning: regularity fights are still going on, and this type is in the cross-fire.  reference_wrapper was forwarded to LWG with an op== that returns bool, and this class does something different.  I have a preference for the design in this paper that forwards along whatever the underlying type does, but be prepared for this to be re-discussed.  That preparation may be to have a link to the minutes were this was discussed in Kona, and sharing that whenever the topic comes up.  Maybe adjust the "Comparisons returning `auto`" section so that it has the word "regular" in it.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -1,4 +1,4 @@
-# Vocabulary Types for Composite Class Design
+# `indirect` and `polymorphic`: Vocabulary Types for Composite Class Design
 
 ISO/IEC JTC1 SC22 WG21 Programming Language C++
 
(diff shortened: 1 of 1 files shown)


jbcoe:
These edits are great Ben, thanks so much.

jbcoe (approved):

jbcoe:
I'm going to merge these and manually fix conflicts in other PRs.

lewissbaker:
> the possible technique of declaring these facilities as allocator-aware containers

One concern here is that, while we want to support the allocator-aware bits, this type does not actually satisfy the requirements of "container" since that implies begin(), end(), iterators, etc.

Is there some other wording for "allocator aware parent object" that doesn't imply "container"?

ben-craig:
I don't think the allocator aware container named requirements directly
uses other container requirements.  I agree that the name is awkward
though.  Perhaps changing the name to allocator aware object throughout the
standard at the same time would remove that concern?

ben-craig:
Allocator aware container _does_ require empty() though, and the authors have gone out of their way to avoid making an empty().

PR #249, opened 2023-11-18
Allocator-construct the owned object

jbcoe:
The tests added in this PR won't pass until #255 is fixed.


jbcoe:
@lewissbaker I'm not sure how well you know allocators but I'm not 100% sure if I need to do this.
I think some work is needed to pass allocators through wrapper types (which the control block is).

jbcoe:
@Ukilele @Twon if folk fancy getting these allocator tests working for the C++14 versions of our types I'd be might grateful.

C++20 gives us the rather magical `make_obj_using_allocator`. C++14 will necessitate writing an equivalent of our own.

jbcoe:
AppleClang14 does not support uninitialized_construct_using_allocator. Looks like we'll need to make our own version.
This will be template-meta-programming-heavy.

PR #252, opened 2023-11-18
Add AppleClang15 to CMake continuous integration tests

jbcoe:
Mostly checking if this will get run at all.


jbcoe:
Looks like Apple Clang 15 is not a thing yet.

PR #260, opened 2023-11-19
Move tracking allocator to a single header to reduce duplication

Twon:


tracking_allocator.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
```
jbcoe, line 1: pragma once won't work properly on some filesystems (distributed cloud stuff, errors are very hard to diagnose). Can we stick to include guards?

jbcoe (dismissed):

tracking_allocator.h
jbcoe: We (I) should set up clang-tidy to pick this up.

jbcoe (approved):

PR #262, opened 2023-11-19
Add C++14 targets for CMake

Twon:


jbcoe (dismissed):
Nice. This is easy to understand now that you've added the target macros too.

jbcoe (approved):

PR #263, opened 2023-11-19
Add clang-tidy checks

jbcoe:


Twon:
Will this apply fixes and recommit?  Presumably, it will just fail and we need to fix it manually?  If so perhaps we should also add the pre-commit bindings?

jbcoe:
This needs a newer version of CMake. Will pick this up again later. It's a distraction for now.

jbcoe:
Just pre-commit is probably more useful. I'll set a local helper script up
first

PR #279, opened 2023-11-20
Bump paper version

Twon:


jbcoe:
I'll defer merging this until we've done a proof read. I can turn off checks and do this locally if that works.

Thanks for the last minute drive to get this ready - looks awesome

jbcoe (approved):

jbcoe (changes requested):
We need to change the date too I think

jbcoe (approved):

PR #292, opened 2023-11-21
Use returns instead of effects where applicable

ben-craig:
Using a more specific element when it makes sense.


Twon (approved):

Twon:
Thanks for your input here @ben-craig!

jbcoe:
Thanks Ben.

PR #300, opened 2023-11-25
Benchmark indirect sorting with and without handling of the valueless state

### LINKED ISSUE #299: Create and profile indirect without preconditions (not valueless) on comparison
We should create versions of indirect without must-not-be-valueless checks on comparison and see how expensive handling the valueless state truly is. Running `std::sort` on a `std::vector` of `indirect<int>` should be a good way to benchmark choices.

jbcoe:
Fixes #299

Build and run the benchmark with: `bazel run -c opt //benchmarks:indirect_sorting_benchmark`


jbcoe:
```
Run on (12 X 24 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x12)
Load Average: 1.67, 1.37, 1.43
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
Indirect_SortingBenchmark              228362681 ns    225794667 ns            3
Indirect_ExperimentalSortingBenchmark  264184021 ns    263328000 ns            2
```

Twon (approved):

PR #302, opened 2023-11-25
Add test to see if indirect is trivially relocatable

jbcoe:


indirect.h
```diff
@@ -47,6 +47,16 @@ namespace xyz {
 }
 #endif  // XYZ_UNREACHABLE_DEFINED
 
+#ifndef XYZ_TRIVIALLY_RELOCATABLE_DEFINED
```
Quuxplusone, line 50:
```suggestion
#ifndef XYZ_TRIVIALLY_RELOCATABLE_DEFINED
#define XYZ_TRIVIALLY_RELOCATABLE_DEFINED
```
(per line 38 above)

indirect_test.cc
```diff
@@ -45,6 +45,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 namespace {
 
+#if XYZ_INDIRECT_IS_TRIVIALLY_RELOCATABLE
```
Quuxplusone, line 48: This has the problem that you're using `std::is_trivially_relocatable_v` but haven't checked `__cpp_lib_trivially_relocatable` yet. I.e., this would fail if compiled on an implementation where the _compiler_ supports `[[trivially_relocatable]]` and `__is_trivially_relocatable` but the _library_ doesn't implement `is_trivially_relocatable_v` and `uninitialized_relocate`.
So I suggest that this should instead be guarded by
```
#if defined(__cpp_lib_trivially_relocatable)
```
and just eliminate the `XYZ_INDIRECT_IS_TRIVIALLY_RELOCATABLE` macro entirely.

indirect.h
```diff
@@ -57,7 +67,7 @@ template <class T, class A>
 inline constexpr bool is_indirect_v<indirect<T, A>> = true;
 
 template <class T, class A = std::allocator<T>>
-class indirect {
+class XYZ_TRIVIALLY_RELOCATABLE indirect {
```
Quuxplusone, line 70: This shouldn't be unconditional; it depends on the allocator. So the building block above should be
```
#define XYZ_TRIVIALLY_RELOCATABLE_IF(x) [[trivially_relocatable(x)]]
```
and then here should be either something simple-and-conservative:
```
class XYZ_TRIVIALLY_RELOCATABLE_IF((std::is_same_v<A, std::allocator<T>>)) indirect {
```
or else something like libc++'s implementation of `vector`:
```
template <class A>
struct allocator_pocma_models_relocatable : std::bool_constant<
    allocator_traits<A>::is_always_equal::value ||
    (allocator_traits<A>::propagate_on_container_copy_assignment::value &&
     allocator_traits<A>::propagate_on_container_move_assignment::value &&
     allocator_traits<A>::propagate_on_container_swap::value)
> {};

template <class A>
inline constexpr bool indirect_be_trivially_relocatable_v =
    std::is_trivially_relocatable_v<A> &&
    std::is_trivially_relocatable_v<typename allocator_traits<A>::pointer> &&
    allocator_pocma_models_relocatable<A>::value;

template<class T, class A = std::allocator<T>>
class XYZ_TRIVIALLY_RELOCATABLE_IF(indirect_be_trivially_relocatable_v<A>) indirect {
```

indirect_test.cc
Quuxplusone: Also, if you have any tests for non-`std::allocator` allocators, it would be good to check here that one of those is _not_ trivially relocatable. For example https://p1144.godbolt.org/z/TTG1r6d71
```
using A = boost::interprocess::allocator<int,
  boost::interprocess::managed_shared_memory::segment_manager>;
static_assert(!std::is_trivially_relocatable_v<xyz::indirect<int, A>>);
```

Quuxplusone (changes requested):
"This shouldn't be unconditional" is important.

jbcoe:
@Quuxplusone Thanks for the great review.

One query: if we make trivially relocatable conditional as you've suggested, I think we lose the support for incomplete types no?

indirect.h
Quuxplusone: Btw, to explain my naming choice here: Because of the precedent set by `<type_traits>`, I think of `foo_is_bar` as _asking a question_ — therefore I needed a different convention for _making an order_. Hence `foo_be_bar` (or sometimes `foo_should_be_bar`, but that's a longer identifier): read "Foo, _be_ bar!" So `foo_be_bar` determines the Platonic barness of foo, which in turn determines the value of `foo_is_bar`.
Not that I'm insisting you should adopt this convention; just explaining in case it looked like a typo. ;)

Quuxplusone:
> One query: if we make trivially relocatable conditional as you've suggested, I think we lose the support for incomplete types no?

I didn't think of that, but I didn't intend to, and I don't _think_ it happens. `xyz::indirect<Incomplete>` should still work fine. It's true that `xyz::indirect<int, IncompleteAllocator<int>>` will not work, but that already didn't work anyway because you need `A` to be complete in order to find out what `typename A::pointer` is.

I encourage you to add test coverage for `xyz::indirect<Incomplete>` if that use-case is important to you!

Twon (dismissed):

Twon (approved):

jbcoe:
I'll close this as we don't want to merge until relocation in C++ is a thing.

Quuxplusone:
> I'll close this as we don't want to merge until relocation in C++ is a thing.

FWIW, as an interested observer, I would prefer to see this taken to completion and merged. It's all properly guarded by the appropriate `#if`s (or at least it should be, and I'm happy to take another look), so it's not _harmful_ to have the codepaths in your main branch (except that it is added complexity, I'd admit). Meanwhile
- it makes your library work "out of the box" with relocatability-enabled compilers, e.g. my Clang fork; you can [test this on Godbolt](https://godbolt.org/z/Ynrs83rzo) or natively
- it serves as an example (for WG21 and for other library maintainers) of how the proposal would work in practice
- it serves as an example (for WG21) that your `indirect`/`polymorphic` proposal doesn't conflict with relocatability

"Merge to trunk, under the proper `#if`s" is also the route taken by e.g. [charles-salvia/std_error](https://github.com/charles-salvia/std_error) (using the feature-test macros) and [STEllAR-GROUP/hpx](https://github.com/STEllAR-GROUP/hpx) (using its own `HPX_HAS_P1144_RELOCATE_AT` macro).

(But it shouldn't be merged _immediately as-is_, because as you can see from Godbolt there's at least two typos in the `#if`'ed-out parts; that's certainly part of the argument in the "complexity" column.)

jbcoe:
Thanks @Quuxplusone . For now the complexity hit is more than I want. I'm trying to minimise cognitive load as we get the paper and reference implementation ready. Feel free to fork.

PR #309, opened 2023-11-28
No need to rebind allocators

jwakely:
There is no reason to mention internal types here. That's needed for node-based containers like `std::set<T>` and `std::map<T>` where we allocate nodes of internal types, and construct a `T`. For those containers we need two different allocator types, so the container stores one allocator type and rebinds to get the other allocator type. That's not relevant for your types, you have no internal node types. You allocate a `T` and construct a `T`, so the `indirect<T>` can just store an `Allocator` and use that directly to both allocate/deallocate and construct/destroy.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -491,11 +491,8 @@ value may only become valueless after it has been moved from.
 object of type `indirect<T, Allocator>` uses an object of type `Allocator` to
 allocate and free storage for the owned object as needed. The owned object is
 constructed using the function
- `allocator_traits<allocator_type>::rebind_traits<U>::construct` and destroyed
-using the function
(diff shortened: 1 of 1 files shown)


Twon (approved):
Thanks for the wording improvements @jwakely!

Twon:
> The same change applies to `polymorphic` too.

I'll ensure this is applied to `polymorphic` too.  Many thanks!

zhihaoy:
> The same change applies to `polymorphic` too.

Doesn't look like so? `polymorphic` allocates control blocks, different from the actual `T`: https://github.com/jbcoe/value_types/blob/main/polymorphic.h#L56-L73

PR #312, opened 2023-11-28
For polymorphic using an allocator with wrong value_type should not be undefined

Twon:


DRAFT.md
```diff
@@ -1123,9 +1122,10 @@ owned object. A polymorphic value may only become valueless after it has been
 moved from.
 
 2. In every specialization `polymorphic<T, Allocator>`, the type
```
jwakely, line 1124:
```suggestion
2. In every specialization `polymorphic<T, Allocator>`, if the type
```

jwakely (approved):

jbcoe (approved):

PR #315, opened 2023-11-28
For polymorphic there is no need to rebind allocators

Twon:


jwakely:
Hmm, maybe I'm wrong about `polymorphic<T>` not requiring any "internal types". Does this type actually allocate an object of some control block type, rather than allocating a `T` directly? I'm pretty sure `indirect<T>` doesn't, that just allocates `T` directly. But the type erasure for `polymorphic` presumably needs to allocate something else. In which case this change would be wrong for `polymorphic`.

jbcoe:
polymorphic uses a control block so I’m pretty sure we need to rebind allocators.

jbcoe (commented):
We need to rebind for the control block.

jbcoe:
We need the rebind for polymorphic otherwise we forbid the use of a control block. I don't see how polymorphic can then be implemented.

Nice work on getting the wording tweaked though - the right edits in the right space. The cost of speculative execution In guess. 😁

PR #320, opened 2023-11-28
Add support for fancy pointers to indirect

jwakely:


jwakely:
It's considerably more work to do this for `indirect` because the code currently traffics in `cblock_t*`, which probably needs to be `std::allocator_traits<allocator_type>::rebind_traits<cblock_t>::pointer` instead.

jbcoe (approved):

PR #321, opened 2023-12-04
Amend T requirements

### LINKED ISSUE #317: Invalid specializations should be ill-formed, not undefined
https://github.com/jbcoe/value_types/blob/4feac3fd38abb31188a8a1e63eef4d724ca026ce/DRAFT.md?plain=1#L524

This means it's undefined behaviour to use `indirect<int&>` or `indirect<void()>`. It should just be a guaranteed error (i.e. ill-formed) not UB.

jbcoe:
Fixes #317

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -522,7 +522,7 @@ move assignment, or swapping of the allocator only if
 
     is `true` within the implementation of the corresponding indirect value operation.
 
-4. The template parameter `T` of `indirect` shall meet the requirements of _Cpp17Destructible_.
+4. A program that instantiates the definition of indirect for a non-object type, an array type, a specialization of indirect, or a cv-qualified type is ill-formed.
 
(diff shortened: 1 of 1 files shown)


Twon (approved):

jwakely (commented):
This outlaws reference types and function types, which is consistent with the previous requirement for _Cpp17Destructible_. It also now forbids `indirect<indirect<X>>` and `indirect<const X>` which were previously allowed. Is that intentional, or just a result of copying the wording from expected/unexpected?

jwakely:
> The meaning of the new wording seems sensible. I see no reason why indirect’s constraints should differ from those of optional. I’m aiming to follow existing art where it makes sense to do so.

`optional<optional<const X>>` is allowed, but this change would forbid `indirect<indirect<const X>>`.

Twon (approved):

PR #335, opened 2024-02-02
Ensure that T in indirect<T> is copy_constructible using constraints

### LINKED ISSUE #313: Default constructor should use Constraints not Mandates
https://github.com/jbcoe/value_types/blob/17eb8a941ceee73a6ea804edbc74bacca99b9738/DRAFT.md?plain=1#L693

The default constructor should be constrained away if the contained value is not default constructible. Otherwise `is_default_constructible_v<indirect<X>>` will always be true, but default constructing an `indirect<X>` might be a hard error.

Make it SFINAE friendly, use _Constraints_.

jbcoe:
Use concept checks on deduced types to delay constraint checks and allow incomplete T for class instantiation.

`indirect<T>` no longer supports `T` where `T` is not copy-constructible.

Fixes #313

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -44,6 +44,8 @@ should not be considered in isolation.
 
 ### Changes in R4
 
+* Use constraints to require that the object owned by `indirect` is copy-constructible.
+
 * Allow comparison of valueless `indirect` objects.
 
 * Remove `std::format` support for `std::indirect`.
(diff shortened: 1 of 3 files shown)


Twon (approved):

Ukilele:
Hey @jbcoe, why does `T` need to be copy-constructible now? I glanced over https://github.com/jbcoe/value_types/issues/313 but I don't see how this issue has any impact on whether `T` should be copy-constructible or not.

PR #342, opened 2024-02-02
Fix effects for constructor wording

### LINKED ISSUE #314: Incorrect effects for default constructor
https://github.com/jbcoe/value_types/blob/17eb8a941ceee73a6ea804edbc74bacca99b9738/DRAFT.md?plain=1#L695-L696

"stores the address" - which address?

Why default constructed not value-initialized?

These effects contradict the [indirect.general] wording which says that the allocator's `construct` function is used. That doesn't necessarily default initialize an object, e.g. for `scoped_allocator` or `polymorphic_allocator` then might propagate the allocator to the `T`.

I think the effects need to say that `alloc` is value-initialized, then describe something like:

```c++
p = alloc.allocate(1);
try {
  allocator_traits<allocator_type>::construct(alloc, to_address(p));
} catch (...) {
  alloc.deallocate(p, 1);
  throw;
}
```

Alternatively, just say it's equivalent to `indirect(allocator_arg, {})` and correctly describe the effects there. (But the `indirect(allocator_arg_t, const Allocator&)` constructor has similar problems, so you need to fix that too, or instead.)

jbcoe:
fixes #314

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -726,8 +726,8 @@ constexpr indirect(const indirect& other);
 
 18. _Mandates_: `T` is a complete type.
 
-19. _Effects_: Equivalent to `indirect(allocator_arg_t{}, Allocator(), other)`
-`
+19. _Effects_: TODO.
+
 20. _Postconditions_: `*this` is not valueless.
 
 ```c++
@@ -743,7 +743,7 @@ constexpr indirect(allocator_arg_t, const Allocator& alloc,
 constexpr indirect(indirect&& other) noexcept;
 ```
 
-23. _Effects_: Equivalent to `indirect(allocator_arg_t{}, Allocator(), std::move(other))`.
+23. _Effects_: TODO.
 
 24. _[Note 1: This constructor does not require that `is_move_constructible_v<T>`
   is `true` --end note]_
@@ -1178,8 +1178,7 @@ constexpr polymorphic(const polymorphic& other);
 
 17. _Mandates_: `T` is a complete type.
 
-18. _Effects_: Equivalent to
-  `polymorphic(allocator_arg_t{}, allocator_traits<allocator_type>::select_on_container_copy_construction(other.alloc), other)`.
+18. _Effects_: TODO.
 
 ```c++
 constexpr polymorphic(allocator_arg_t, const Allocator& a,
@@ -1194,7 +1193,7 @@ constexpr polymorphic(allocator_arg_t, const Allocator& a,
 constexpr polymorphic(polymorphic&& other) noexcept;
 ```
 
-21. _Effects_: Equivalent to `polymorphic(allocator_arg_t{}, Allocator(other.alloc_), other)`.
+21. _Effects_: TODO.
 
 ```c++
 constexpr polymorphic(allocator_arg_t, const Allocator& a,
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -726,8 +726,8 @@ constexpr indirect(const indirect& other);
 
 18. _Mandates_: `T` is a complete type.
 
-19. _Effects_: TODO.
-
+19. _Effects_: Equivalent to `indirect(allocator_arg_t{}, Allocator(), other)`
+`
 20. _Postconditions_: `*this` is not valueless.
 
 ```c++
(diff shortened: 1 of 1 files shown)


DRAFT.md
```diff
@@ -1175,7 +1168,7 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a,
 
 14. _Mandates_: `T` is a complete type.
 
-15. _Effects_: `alloc` is direct-non-list-initialized with `a`.
+15. _Effects_: TODO.
```
Twon, line 1171:
```suggestion
15. _Effects_: `polymorphic(allocator_arg_t{}, Allocator(), other)`.
```

DRAFT.md
```diff
@@ -665,8 +665,7 @@ explicit constexpr indirect()
 
 2. _Mandates_: `T` is a complete type.
 
-3. _Effects_: Constructs an `indirect` owning a default constructed `T`
-  and stores the address in `p`. `alloc` is default constructed.
+3. _Effects_: Equivalent to `indirect(allocator_arg_t{}, Allocator())`.
```
Twon, line 668: `std::vector` has a constructor taking an allocator (but without the `std::allocator_arg_t` parameter: https://eel.is/c++draft/vector#cons

```
constexpr explicit vector(const Allocator&) noexcept;
[1](https://eel.is/c++draft/vector#cons-1)
[#](http://github.com/Eelis/draft/tree/7faaaf22a11d54ad00e40cc8d7fbee90db8b0837/source/containers.tex#L8773)
Effects: Constructs an empty vector, using the specified allocator[.](https://eel.is/c++draft/vector#cons-1.sentence-1)
[2](https://eel.is/c++draft/vector#cons-2)
[#](http://github.com/Eelis/draft/tree/7faaaf22a11d54ad00e40cc8d7fbee90db8b0837/source/containers.tex#L8778)
Complexity: Constant[.](https://eel.is/c++draft/vector#cons-2.sentence-1)
```

DRAFT.md
```diff
@@ -682,8 +681,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& alloc);
 
 7. _Mandates_: `T` is a complete type.
 
-8. _Effects_: Constructs an `indirect` owning a default constructed `T` and
-  stores the address in `p`. `alloc` is direct-non-list-initialized with `alloc`.
+8. _Effects_: TODO.
```
Twon, line 684: `std::vector` has a constructor taking an allocator (but without the `std::allocator_arg_t` parameter: https://eel.is/c++draft/vector#cons

```
constexpr explicit vector(const Allocator&) noexcept;
[1] Effects: Constructs an empty vector, using the specified allocator.
[2] Complexity: Constant.
```

DRAFT.md
```diff
@@ -740,15 +737,13 @@ constexpr indirect(allocator_arg_t, const Allocator& alloc,
...
+22. _Effects_: TODO.
 
 ```c++
 constexpr indirect(indirect&& other) noexcept;
 ```
 
-23. _Effects_: Constructs an `indirect` that takes ownership of the `other`'s
-    owned object and stores the address in `p_`. `allocator_` is initialized by
-    construction from `other.allocator_`.
+23. _Effects_: Equivalent to `indirect(allocator_arg_t{}, Allocator(), std::move(other))`.
```
Twon, line 746: Here is how unique_ptr handles specifying the effect via the post condition here: https://eel.is/c++draft/unique.ptr.single.ctor:

```
constexpr unique_ptr(unique_ptr&& u) noexcept;
[15] Constraints: is_move_constructible_v<D> is true
[16] Preconditions: If D is not a reference type, D meets the Cpp17MoveConstructible requirements (Construction of the deleter from an rvalue of type D does not throw an exception[.](https://eel.is/c++draft/unique.ptr.single.ctor#16.sentence-2)
[17] Effects: Constructs a unique_ptr from u. If D is a reference type, this deleter is copy constructed from u's deleter; otherwise, this deleter is move constructed from u's deleter 
[Note: The construction of the deleter can be implemented with std​::​forward<D> — end note]
[18] Postconditions: get() yields the value u.get() yielded before the construction u.get() == nullptr get_deleter() returns a reference to the stored deleter that was constructed from u.get_deleter() If D is a reference type then get_deleter() and u.get_deleter() both reference the same lvalue deleter
```

jbcoe: Nice. I reckon that's half of the story. Allocators might complicate matters.

DRAFT.md
jbcoe: "using the specified allocator" is the phrase-of-power that we need. Great spot!

Twon (approved):

Twon (approved):

PR #344, opened 2024-02-02
Fix issues using "polymorphic_cxx14.h" in C++17 and later

Quuxplusone:
The non-C++14 `polymorphic` requires C++20.
If the C++14 `polymorphic` is being used in C++17 mode, then both `xyz::in_place_type_t` and `std::in_place_type_t` will exist. Make sure they have the same definition in all places. This fixes a compilation error when compiling "polymorphic_test.h" with `-std=c++17 -DXYZ_POLYMORPHIC_CXX_14`.

Also, match the open/close of `namespace xyz` with the open/close of preprocessor conditionals, so that you don't get a compilation error when compiling "indirect.h" followed by "polymorphic_cxx14.h" in `-std=c++20` mode.


Twon (approved):
Thanks for raising this.  I've added C++ 17 test for CMake and Bazel to ensure C++17 issues do not happen again: https://github.com/jbcoe/value_types/pull/346

PR #345, opened 2024-02-02
[do not merge] Implement P1144 trivially-relocatable for `indirect` and `polymorphic`

Quuxplusone:
This is my take on #302, just to have a "clean copy." I would _like_ it merged, but I _expect_ it to be closed, per the discussion on #302 about not wanting to maintain the extra complexity. (My goal is just to record this branch history in a neat place for any posterity who want to [play with it on Godbolt.](https://godbolt.org/z/rPoK4PT4b))

I've added the same logic for `polymorphic<T>`. (#302 did it only for `indirect<T>`.)
While doing this, I noticed:
- Neither type has any unit tests for `is_nothrow_move_constructible`, `!is_trivially_fooable`, etc. I think such unit tests should be added.
- `indirect<T>` correctly uses `allocator_traits<A>::pointer`, but `polymorphic<T>` hard-codes `cblock_t*` instead of `allocator_traits<A>::rebind_alloc<cblock_t>::pointer`. This means `polymorphic<T>` isn't compatible with Boost.Interprocess `offset_ptr` allocators. I could work up such a PR if you're interested; but (next bullet)
- I'm not sure I understand all the various versions of polymorphic that seem to exist side-by-side right now.
- I see a lot of uses of `XYZ_HAS_STD_OPTIONAL` (invariably true in C++17 and later) in "indirect_test.cc" (requires C++20 or later); these could perhaps be removed, unless the plan is to backport `indirect` to C++14-or-earlier just like `polymorphic` was backported.

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -49,7 +49,8 @@ namespace xyz {
 
 #ifndef XYZ_TRIVIALLY_RELOCATABLE_IF_DEFINED
 #define XYZ_TRIVIALLY_RELOCATABLE_IF_DEFINED
-#if defined(__cpp_impl_trivially_relocatable) && defined(__cpp_lib_trivially_relocatable)
+#if defined(__cpp_impl_trivially_relocatable) && \
+    defined(__cpp_lib_trivially_relocatable)
 #define XYZ_TRIVIALLY_RELOCATABLE_IF(x) [[trivially_relocatable(x)]]
 #else
 #define XYZ_TRIVIALLY_RELOCATABLE_IF(x)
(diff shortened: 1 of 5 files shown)


Twon:
Added an issue to track adding tests bases on `offset_ptr` to ensure `allocator_traits<A>::pointer` is correctly supported.

Twon:
Added an issue to track adding tests bases on `offset_ptr` to ensure `allocator_traits<A>::pointer` is correctly supported: https://github.com/jbcoe/value_types/issues/348

Twon:
Regarding the comment on various versions of polymorphic...

> `I'm not sure I understand all the various versions of polymorphic that seem to exist side-by-side right now.`

What can we do to make this clearer? Should a note be added to the readme on the different implementations?

Basically, there are 4 versions:

- C++20 implementation of polymorphic
- C++14/17 compatible implementation of polymorphic for backwards compatibility
- An implementation of polymorphic using the small buffer optimisation
- An implementation of polymorphic using inline vtable

The 2 versions in experimentation were provided because questions we made about whether either of these approaches should be preferred during discussion of the designs at LEWG. The intention in supporting these is the design space is fully explored.

Please let me know if you have any questions or if there are any actions we can take to better document the purpose of supporting these.

Quuxplusone:
@jbcoe @Twon fyi, https://github.com/llvm/llvm-project/pull/84621 is in review right now, to implement `__is_trivially_relocatable(T)` in Clang in a way that conforms to P1144 and is suitable for use in `std::is_trivially_relocatable_v` as used in this patch. (No attribute yet, but that might be coming next, at which point this patch would be completely implementable in mainline Clang!) I know trivial relocation is just a very small tangent to this repo in general; but I'd love it if you were interested enough to visit that PR and say "yeah this would be useful for our purposes" or "no it wouldn't."

jbcoe:
Closing as P1144 is not the direction chosen by the C++ standards committee.

PR #350, opened 2024-02-03
Clean up constructor wording

jbcoe:


DRAFT.md
```diff
@@ -726,45 +726,40 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a, U&& u, Us&& ...
 constexpr indirect(const indirect& other);
 ```
 
-18. _Mandates_: `T` is a complete type.
```
Twon, line 729: Why is this no longer true?  I would expect `T` must be complete so we can call the underlying copy constructor

Twon (approved):
Minor query, but otherwise looks good.

DRAFT.md
jbcoe: My reasoning was that we're not imposing any constraints on `T`. The copy could be done in another TU and linked in.

PR #351, opened 2024-02-03
Fixes for move constructors to handle non-equal allocators

### LINKED ISSUE #349: Fix wording for move constructors when `allocator_traits::is_always_equal` is false
If `allocator_traits::is_always_equal` is `false` then there may be two allocators allocating from different memory domains so you cannot move from one to the other.  In this case, you must allocate a new object from the object's primary allocator and then move from the right-hand side object into the newly constructed object. As such the type cannot be incomplete here as you need the full definition in order to use the move constructor, as opposed to swapping pointers (where the type can be incomplete) when `allocator_traits::is_always_equal` is `true`

jbcoe:
Fixes #349

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -12,16 +12,27 @@ cc_library(
     visibility = ["//visibility:public"],
 )
 
+cc_library(
+    name = "tagged_allocator",
+    hdrs = ["tagged_allocator.h"],
+)
+
+cc_library(
+    name = "tracking_allocator",
+    hdrs = ["tracking_allocator.h"],
+)
+
 cc_test(
     name = "indirect_test",
     size = "small",
     srcs = [
         "indirect_test.cc",
-        "tracking_allocator.h"
     ],
     deps = [
         "feature_check",
         "indirect",
+        "tagged_allocator",
+        "tracking_allocator",
         "@com_google_googletest//:gtest_main",
     ],
 )
@@ -29,8 +40,8 @@ cc_test(
 cc_library(
     name = "indirect_cxx14",
     hdrs = [
-        "indirect_cxx14.h",
         "indirect.h",
+        "indirect_cxx14.h",
     ],
     copts = ["-Iexternal/value_types/"],
     defines = ["XYZ_INDIRECT_CXX_14"],
@@ -42,14 +53,15 @@ cc_test(
     size = "small",
     srcs = [
         "indirect_test.cc",
-        "tracking_allocator.h"
     ],
+    copts = ["-std=c++14"],
     deps = [
         "feature_check",
         "indirect_cxx14",
+        "tagged_allocator",
+        "tracking_allocator",
         "@com_google_googletest//:gtest_main",
     ],
-    copts = ["-std=c++14"],
 )
 
 cc_test(
@@ -57,14 +69,15 @@ cc_test(
     size = "small",
     srcs = [
         "indirect_test.cc",
-        "tracking_allocator.h"
     ],
+    copts = ["-std=c++17"],
     deps = [
         "feature_check",
         "indirect_cxx14",
+        "tagged_allocator",
+        "tracking_allocator",
         "@com_google_googletest//:gtest_main",
     ],
-    copts = ["-std=c++17"],
 )
 
 cc_library(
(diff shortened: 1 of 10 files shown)


BUILD.bazel
```diff
@@ -12,26 +12,42 @@ cc_library(
     visibility = ["//visibility:public"],
 )
 
+cc_library(
```
Twon, line 15: What is the logic for mapping a single file per library?  Seems these should all be part of a common "testing helpers" library but there may be something of the set-up Bazel I don't understand so wanted to clarify.

DRAFT.md
```diff
@@ -1208,16 +1206,15 @@ constexpr polymorphic(allocator_arg_t, const Allocator& a,
...
-    the owned value of `other` using the specified allocator, considering that
-    owned value as an rvalue.
-
-22. _[Note 1: This constructor does not require that `is_move_constructible_v<T>`
-  is `true`. --end note]_
-
-23. _[Note 2: The use of this function may require that `T` be a complete type
+    `other` is valueless, `*this` is valueless. Otherwise, if `alloc == other.alloc`
+    either constructs an object of type `polymorphic` that owns the owned value
+    of other, making`other` is valueless; or, owns an object of the same type
```
Twon, line 1211: Firstly looks like a space is missing, but also this does not ready quite right.  Should "making `other` is valueless" be "making `other` valueless" or "`other` is valueless" 
```suggestion
    of other, making `other` is valueless; or, owns an object of the same type
```

Twon: Having re-read multiple times, I think this should be:
```suggestion
    of other, making `other` valueless; or, owns an object of the same type
```

DRAFT.md
```diff
@@ -1208,16 +1206,15 @@ constexpr polymorphic(allocator_arg_t, const Allocator& a,
...
+    constructed from the owned value of `other` using the specified allocator,
+    considering that owned value as an rvalue. Otherwise constructs an object of
```
Twon, line 1213: I think this says the same thing twice...

> Otherwise constructs an object of type `polymorphic` using the specified allocator, considering that owned value as an rvalue.

How is this different from "owns an object of the same type constructed from the owned value of `other` using the specified allocator, considering that owned value as an rvalue." as I think this covers the case where allocators differ and the memory must be allocated from the new allocator and then constructed with `T`'s move constructor?

indirect_cxx14.h
```diff
@@ -179,16 +166,33 @@ class indirect : private detail::empty_base_optimization<A> {
...
+        if (!other.valueless_after_move()) {
+          p_ = construct_from(alloc_base::get(), std::move(*other));
+        } else {
+          p_ = nullptr;
+        }
+      }
+    }
   }
 
+  indirect(indirect&& other) noexcept
```
Twon, line 192:
```suggestion
  indirect(indirect&& other) noexcept(allocator_traits::is_always_equal::value)
```

indirect_cxx14.h
```diff
@@ -179,16 +166,33 @@ class indirect : private detail::empty_base_optimization<A> {
...
-      : alloc_base(other.alloc_base::get()), p_(nullptr) {
-    std::swap(p_, other.p_);
-  }
+  indirect(const indirect& other)
+      : indirect(std::allocator_arg,
+                 allocator_traits::select_on_container_copy_construction(
+                     other.alloc_base::get()),
+                 other) {}
 
   indirect(std::allocator_arg_t, const A& alloc, indirect&& other) noexcept
```
Twon, line 175:
```suggestion
  indirect(std::allocator_arg_t, const A& alloc, indirect&& other) noexcept(allocator_traits::is_always_equal::value)
```

Twon (changes requested):
A few minor point to address

BUILD.bazel
jbcoe: There's no need for a library per file. Your suggestions works too.

DRAFT.md
jbcoe: Let's discuss offline

jbcoe: Repetition is intended. Added a drafting note.

Twon (approved):

PR #354, opened 2024-02-05
[OLD] Update ctors for exception safety

jbcoe:
If folk can review the implementation I'll move onto wording once we're happy.

Ideally we want tests to check that the strong-exception-safety guarantee holds.

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -165,14 +165,16 @@ class indirect {
     } else {
       if (std::is_copy_assignable_v<T> && !valueless_after_move() &&
           alloc_ == other.alloc_) {
-        *p_ = *other;
+        T tmp(*other);
+        using std::swap;
+        swap(tmp, *p_);
       } else {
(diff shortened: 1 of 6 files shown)


indirect.h
```diff
@@ -178,21 +188,31 @@ class indirect {
...
-        reset();  // using current allocator.
-        alloc_ = other.alloc_;
-      }
-    }
-    reset();  // We may not have reset above and it's a no-op if valueless.
+    // Check to see if the allocators need to be updated.
+    // We defer actually updating the allocator until later because it may be
+    // needed to delete the current control block.
+    bool update_alloc =
+        allocator_traits::propagate_on_container_copy_assignment::value &&
```
Twon, line 195: I think this should be `propagate_on_container_move_assignment` and not `propagate_on_container_copy_assignment` right?  Looks like that is what is was previously...
```suggestion
        allocator_traits::propagate_on_container_move_assignment::value &&
```

PR #364, opened 2024-02-07
Remove comparsion ops where they can be compiler-synthesized

jbcoe:


Twon (approved):

Twon:
We are seeing intermittent build failures due to the following error:
```
dyld[4315]: Symbol not found: __ZNSt3__13pmr15memory_resourceD2Ev
```

[This issue](https://bugreports.qt.io/browse/QTBUG-114316) suggests that it was caused by building on macOS 13.5.2 with the macOS 14.0 SDK... I'm not yet sure what the fix should be or if this is caused by the GitHub action mac image in use.

PR #370, opened 2024-02-14
`indirect` and `polymorphic` to support converting constructors

Twon:


DRAFT.md
```diff
@@ -1554,24 +1599,21 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
-### Implicit conversions
+### Converting constructors
 
-We decided that there should be no implicit conversion of a value `T` to an
-`indirect<T>` or `polymorphic<T>`. An implicit conversion would require using
-the free store and memory allocation, which is best made explicit by the user.
+After reciving feedback from LEWG, converting constructors were added to
+support conversion from `T` to an `indirect<T>` or `polymorphic<T>`. Because
+these operation allocate they memory are marked explicit so it the intent is
+clear.
```
jbcoe, line 1607: I think this should say “marked explicit so that the intent to use them made clear” or something like that.

jbcoe (approved):

DRAFT.md
```diff
@@ -743,41 +757,55 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 10. _Throws_: Nothing unless `allocator_traits<allocator_type>::allocate` or
   `allocator_traits<allocator_type>::construct` throws.
 
+```c++
+template <class U>
+explicit constexpr indirect(U&& u);
+```
+
+11. _Constraints_: `is_constructible_v<T, U&&>` is `true`.
```
Twon, line 765: I think this constraint is wrong, as this will result in a difference between polymorphic and indirect behaviour.  This should restrict U to be `is_same<T, remove_cvref<U>>`

jbcoe (dismissed):

PR #372, opened 2024-02-14
`indirect` and `polymorphic` converting assignment operators

Twon:


polymorphic.h
```diff
@@ -226,6 +226,21 @@ class polymorphic {
...
+  constexpr polymorphic& operator=(U&& u)
+    requires(not std::same_as<polymorphic, std::remove_cvref_t<U>>) &&
+            std::is_copy_assignable_v<std::remove_cvref_t<U>> &&
+            std::derived_from<std::remove_cvref_t<U>, T>
+  {
+    if (valueless_after_move()) {
+      *this = polymorphic<T>(std::in_place_type_t<std::remove_cvref_t<U>>{},
+                             std::forward<U>(u));
+    } else {
+      this->operator*() = std::forward<U>(u);
```
Twon, line 239: @jbcoe This is insufficient for an assignment to a polymorphic because you don't know that the types on lhs, and rhs match.  So the only sensible thing to do seems to be to recreate the object from scratch?

PR #380, opened 2024-03-11
Update assignment wording

### LINKED ISSUE #377: Update assignment wording
Consider wording from https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2047r7.html for assignment.

jbcoe:
Fixes #377

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -46,9 +46,6 @@ should not be considered in isolation.
 
 * Improve wording for assignment operators to remove ambiguity.
 
-* Remove postcondition that an indirect or polymorphic object is valueless
-  after being assigned from.
-
 ### Changes in R6
 
 * Add `std::in_place_t` argument to indirect constructors.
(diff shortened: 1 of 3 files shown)


indirect_cxx14.h
```diff
@@ -197,9 +197,7 @@ class indirect : private detail::empty_base_optimization<A> {
     } else {
       if (std::is_copy_assignable<T>::value && !valueless_after_move() &&
           alloc_base::get() == other.alloc_base::get()) {
-        T tmp(*other);
```
jbcoe, line 200: We assign rather than copy-swap this because we _just_ want to store an object in dynamically assigned memory not alter exception safety guarantees. basic_optional in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2047r7.html does the same.

indirect.h
```diff
@@ -158,9 +158,7 @@ class indirect {
     } else {
       if (std::is_copy_assignable_v<T> && !valueless_after_move() &&
           alloc_ == other.alloc_) {
-        T tmp(*other);
```
jbcoe, line 161: We assign rather than copy-swap this because we just want to store an object in dynamically assigned memory not alter exception safety guarantees. basic_optional in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2047r7.html does the same.

nbx8 (approved):
Wording clarity has improved.

DRAFT.md
```diff
@@ -842,25 +846,30 @@ constexpr indirect& operator=(const indirect& other);
...
+  previously owned object in this, if any, is destroyed using
   `allocator_traits<allocator_type>::destroy` and then deallocated using
   `allocator_traits<allocator_type>::deallocate`.
 
   If the allocator needs updating, the allocator in `this` is replaced with a
   copy of the allocator in `other`.
 
-  No effects if an exception is thrown.
-
 3. _Returns_: A reference to `*this`.
```
Twon, line 865: The following sequence of numbers needs to be updated following  the addition if this statement

Twon (changes requested):
An update to the numbering around the additional statement is required (added a comment to highlight this)

DRAFT.md
jbcoe: Well-spotted! Fixed.

Twon (approved):

PR #396, opened 2024-03-17
Discuss conditional copy construction with new template trickery

jbcoe:


DRAFT.md
```diff
@@ -1560,27 +1560,21 @@ Both `indirect` and `polymorphic` support incomplete types. Support for an
 incomplete type requires deferring the instantiation of functions with
 requirements until they are used.
 
-For example, the default constructor of `indirect` requires that `T` is default
+The default constructor of `indirect` requires that `T` is default
 constructible. We can't write this constraint as a requirement on `T` because
```
nbx8, line 1564: We cannot write this constraint...

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
 and move constructible. The authors do not envisage that this could be relaxed
 in a future version of the C++ standard.
 
+While a copy constructor cannot be a template, we can conditionally constrain
+copy construction of `indirect`, in C++20 and later, by defining:
```
nbx8, line 1587: While a copy constructor cannot be a template, in C++20 and later we can conditionally constrain
copy construction of `indirect` by defining:

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
+
+```c++
+indirect(const indirect& other) requires false = delete;
+
+template <typename TT=T>
```
nbx8, line 1592: template <typename TT = T>

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
+indirect(const indirect& other) requires is_copy_constructible_v<TT>;
+```
+
+An instantiation of the function template, with `TT=T`, is added to the overload
```
nbx8, line 1596: An instantiation of the function template with `TT=T` is added ...

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
+set when `indirect` is copy-constructed and will be selected if the owned
+object type `T` is copy-constructible.
+
+This would make copy construction conditional for `indirect` but not for
```
nbx8, line 1600: Suggestion: make this one paragraph ?

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
+`polymorphic`. We opt for consistency and make copy construction unconditional
+for both `indirect` and `polymorphic`. Making `indirect` conditionally copy
+constructible in a future version of the C++ standard would require adding a
```
nbx8, line 1603: copy-constructible

DRAFT.md
```diff
@@ -1589,6 +1583,28 @@ such types are copy constructible. We make `polymorphic` unconditionally copy
...
+template function as above and would be an ABI break. It might be simpler to add
+new types for non-copyable `indirect` and `polymorphic` objects although we do
```
nbx8, line 1605: objects, although we do not...

nbx8 (approved):
Minor edits suggested, otherwise all good to go.

DRAFT.md
jbcoe: I'm not sure what you mean? I've possibly missed the same typo twice in my head.

`indirect(const indirect& other) requires false = delete;` is not a function template it's a non-template function from a class template (which means we can constrain it).

DRAFT.md
jbcoe: I've gone for 'copy constructible' and de-hyphenated everywhere. US folk tend not to use hy-phens.

DRAFT.md
jbcoe: Missing spacing discussed locally and fixed.

PR #398, opened 2024-03-20
Amend wording after discussion with Jonathan Wakely

jbcoe:
We already mention allocators in:  Class template polymorphic general [polymorphic.general] 

_In every specialization `polymorphic<T, Allocator>`, if the type
`allocator_traits<Allocator>::value_type` is not the same type as`T`, the
program is ill-formed. Every object of type `polymorphic<T, Allocator>` uses an
object of type `Allocator` to allocate and free storage for the owned object as
needed. The owned object is constructed using the function
`allocator_traits<allocator_type>::rebind_traits<U>::construct` and destroyed
 using the function
`allocator_traits<allocator_type>::rebind_traits<U>::destroy`, where `U` is
either `allocator_type::value_type` or an internal type used by the polymorphic
value._

There is, perhaps, no need to mention allocators again for individual constructors.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -591,13 +591,15 @@ become valueless after it has been moved from.
 `allocator_traits<Allocator>::value_type` is not the same type as `T`, the
 program is ill-formed. Every object of type `indirect<T, Allocator>` uses an
 object of type `Allocator` to allocate and free storage for the owned object as
-needed. The owned object is constructed using the function
-`allocator_traits<allocator_type>::construct` and destroyed using the function
-`allocator_traits<allocator_type>::destroy`.
+needed.
+
+3. Constructing an owned object using an allocator `a` and arguments `args...`
+means calling `allocator_traits<allocator_type>::construct(a, p, args...)` where
+`p` is a pointer to suitable storage.
 
 // DRAFTING NOTE: [indirect.general]#3 modeled on [container.reqmts]#64
 
-3. Copy constructors for an indirect value obtain an allocator by calling
+4. Copy constructors for an indirect value obtain an allocator by calling
 `allocator_traits<allocator_type>::select_on_container_copy_construction` on the
 allocator belonging to the indirect value being copied. Move constructors obtain
 an allocator by move construction from the allocator belonging to the object
@@ -622,15 +624,15 @@ move assignment, or swapping of the allocator only if
     is `true` within the implementation of the corresponding indirect value
     operation.
 
-4. A program that instantiates the definition of indirect for a non-object type,
+5. A program that instantiates the definition of indirect for a non-object type,
    an array type, or a cv-qualified type is ill-formed.
 
-5. The template parameter `T` of `indirect` may be an incomplete type.
+6. The template parameter `T` of `indirect` may be an incomplete type.
 
-6. The template parameter `Allocator` of `indirect` shall meet the
+7. The template parameter `Allocator` of `indirect` shall meet the
    _Cpp17Allocator_ requirements.
 
-7. If a program declares an explicit or partial specialization of `indirect`,
+8. If a program declares an explicit or partial specialization of `indirect`,
    the behavior is undefined.
 
 #### X.Y.2 Class template indirect synopsis [indirect.syn]
@@ -1102,14 +1104,15 @@ object may only become valueless after it has been moved from.
 `allocator_traits<Allocator>::value_type` is not the same type as`T`, the
 program is ill-formed. Every object of type `polymorphic<T, Allocator>` uses an
 object of type `Allocator` to allocate and free storage for the owned object as
-needed. The owned object is constructed using the function
-`allocator_traits<allocator_type>::rebind_traits<U>::construct` and destroyed
- using the function
-`allocator_traits<allocator_type>::rebind_traits<U>::destroy`, where `U` is
+needed.
+
+3. Constructing an owned object using an allocator `a` and arguments
+`args...` means calling `allocator_traits<allocator_type>::rebind_traits<U>::construct(a, p,
+args...)` where `p` is a pointer to suitable storage and `U` is
 either `allocator_type::value_type` or an internal type used by the polymorphic
 value.
 
-3. Copy constructors for a polymorphic value obtain an allocator by calling
+4. Copy constructors for a polymorphic value obtain an allocator by calling
 `allocator_traits<allocator_type>::select_on_container_copy_construction` on the
 allocator belonging to the polymorphic value being copied. Move constructors
 obtain an allocator by move construction from the allocator belonging to the
@@ -1129,15 +1132,15 @@ or (64.3) `allocator_traits<allocator_type>::propagate_on_container_swap::value`
 is true within the implementation of the corresponding polymorphic value
 operation.
 
-4. A program that instantiates the definition of polymorphic for a non-object
+5. A program that instantiates the definition of polymorphic for a non-object
    type, an array type, or a cv-qualified type is ill-formed.
 
-5. The template parameter `T` of `polymorphic` may be an incomplete type.
+6. The template parameter `T` of `polymorphic` may be an incomplete type.
 
-6. The template parameter `Allocator` of `polymorphic` shall meet the
+7. The template parameter `Allocator` of `polymorphic` shall meet the
    requirements of _Cpp17Allocator_.
 
-7. If a program declares an explicit or partial specialization of `polymorphic`,
+8. If a program declares an explicit or partial specialization of `polymorphic`,
    the behavior is undefined.
 
 #### X.Z.2 Class template polymorphic synopsis [polymorphic.syn]
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -44,6 +44,12 @@ should not be considered in isolation.
 
 ### Changes in R7
 
+* Add more explicit wording for use of `allocator_traits::construct` in
+  `indirect` and `polymorphic` constructors.
+
+* Discuss `indirects` non-conditional copy constructior in the light of
+  implementation tricks that would enable it.
+
 * Improve wording for assignment operators to remove ambiguity.
 
 * Add motivation for `valueless_after_move` member function.
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -1108,8 +1108,8 @@ needed.
 
 3. Constructing an owned object using an allocator `a` and arguments `args...`
 means calling `allocator_traits<allocator_type>::rebind_traits<U>::construct(a,
-p, args...)` where `p is a pointer obtained by calling
-allocator_traits<allocator_type>::rebind_traits<U>::allocate` and `U` is either
+p, args...)` where `p` is a pointer obtained by calling
+`allocator_traits<allocator_type>::rebind_traits<U>::allocate` and `U` is either
 `allocator_type::value_type` or an internal type used by the polymorphic value.
 
 4. Copy constructors for a polymorphic value obtain an allocator by calling
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -873,7 +873,7 @@ constexpr indirect& operator=(const indirect& other);
   object is assigned to `*other`.
 
   Otherwise, if `alloc != other.alloc` or `this` is valueless, a new owned
-  object is constructed in `this` using
+  object is constructed in `*this` using
   `allocator_traits<allocator_type>::construct` with the owned object from
   `other` as the argument, using either the allocator in `this` or the allocator
   in `other` if the allocator needs updating. The previously owned object in
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -906,7 +906,7 @@ constexpr indirect& operator=(indirect&& other) noexcept(
   `*this`, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
   and then the storage is deallocated.
 
-  Otherwise, if `alloc == other.alloc`, swaps the owned objects in `this` and
+  Otherwise, if `alloc == other.alloc`, swaps the owned objects in `*this` and
   `other`; the owned object in `other`, if any, is then destroyed using
   `allocator_traits<allocator_type>::destroy` and then the storage is deallocated.
 
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -987,7 +987,7 @@ constexpr allocator_type get_allocator() const noexcept;
 ```c++
 constexpr void swap(indirect& other) noexcept(
   allocator_traits::propagate_on_container_swap::value
-  || allocator_traits::is_always_equal::value);
+  || allocator_traits<Allocator>::is_always_equal::value);
 ```
 
 1. _Effects_: Swaps the states of `*this` and `other`, exchanging owned objects
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -1387,7 +1387,7 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
   storage is deallocated.
 
   Otherwise, if `alloc != other.alloc`; if `other` is not valueless, a new owned
-  object is constructed in `this` using
+  object is constructed in `*this` using
   `allocator_traits<allocator_type>::rebind_traits::construct` with the owned
   object from `other` as the argument as an rvalue, using either the allocator
   in `*this` or the allocator in `other` if the allocator needs updating. The
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -866,7 +866,7 @@ constexpr indirect& operator=(const indirect& other);
   `true` and `alloc != other.alloc` then the allocator needs updating.
 
   If `other` is valueless, `*this` becomes valueless and the owned object in
-  this, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
+  `*this`, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
   and then the storage is deallocated.
 
   Otherwise, if `alloc == other.alloc` and `this` is not valueless, the owned
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -917,7 +917,7 @@ constexpr indirect& operator=(indirect&& other) noexcept(
   the allocator in `other` if the allocator needs updating. The previous owned
   object in `*this`, if any, is destroyed using
   `allocator_traits<allocator_type>::destroy` and then the storage is
-  deallocated. If the allocator needs updating, the allocator in `this` is
+  deallocated. If the allocator needs updating, the allocator in `*this` is
   replaced with a copy of the allocator in `other`.
 
 6. _Postconditions_: `other` is valueless.
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -751,7 +751,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 7. _Mandates_: `T` is a complete type.
 
 8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs owned
-   object of type `T` with no arguments.
+   object of type `T` with an empty argument list.
 
 9. _Postconditions_: `*this` is not valueless.
 
@@ -1227,7 +1227,7 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a);
 7. _Mandates_: `T` is a complete type.
 
 8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs an
-   owned object of type `T` with no arguments.
+   owned object of type `T` with an empty argument list.
 
 9. _Postconditions_: `*this` is not valueless.
 
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -1390,7 +1390,7 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
   object is constructed in `this` using
   `allocator_traits<allocator_type>::rebind_traits::construct` with the owned
   object from `other` as the argument as an rvalue, using either the allocator
-  in `this` or the allocator in `other` if the allocator needs updating. The
+  in `*this` or the allocator in `other` if the allocator needs updating. The
   previous owned object in `*this`, if any, is destroyed using
   `allocator_traits<allocator_type>::rebind_traits::destroy` and then the
   storage is deallocated.
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -877,7 +877,7 @@ constexpr indirect& operator=(const indirect& other);
   `allocator_traits<allocator_type>::construct` with the owned object from
   `other` as the argument, using either the allocator in `this` or the allocator
   in `other` if the allocator needs updating. The previously owned object in
-  this, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
+  `*this`, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
   and then the storage is deallocated.
 
   If the allocator needs updating, the allocator in `this` is replaced with a
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -49,7 +49,7 @@ should not be considered in isolation.
 
 ### Changes in R7
 
-* Discuss `indirects` non-conditional copy constructior in the light of
+* Discuss `indirect`'s non-conditional copy constructor in the light of
   implementation tricks that would enable it.
 
 * Improve wording for assignment operators to remove ambiguity.
@@ -601,7 +601,7 @@ program is ill-formed. Every object of type `indirect<T, Allocator>` uses an
 object of type `Allocator` to allocate and free storage for the owned object as
 needed.
 
-3. Constructing an owned object using an allocator `a` and arguments `args...`
+3. Constructing an owned object with arguments `args...` using an allocator `a`
 means calling `allocator_traits<allocator_type>::construct(a, p, args...)` where
 `p` is a pointer obtained by calling
 `allocator_traits<allocator_type>::allocate`.
@@ -759,7 +759,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 
 7. _Mandates_: `T` is a complete type.
 
-8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs owned
+8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs an owned
    object of type `T` with an empty argument list, using the allocator `alloc`.
 
 9. _Postconditions_: `*this` is not valueless.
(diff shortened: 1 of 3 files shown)


Twon (approved):

DRAFT.md
```diff
@@ -823,8 +821,8 @@ constexpr indirect(allocator_arg_t, const Allocator& a, indirect&& other)
 22. _Effects_: `alloc` is direct-non-list-initialized with `a`. If
     `other` is valueless, `*this` is valueless. Otherwise, if `alloc ==
     other.alloc` constructs an object of type `indirect` that owns the owned
-    value of other; `other` is valueless. Otherwise constructs an object of type
-    `indirect` using the specified allocator with `*other` used as an rvalue.
+    value of other; `other` is valueless. Otherwise constructs an owned object of type
+    `T` with `*other` used as an rvalue.
```
jwakely, line 825:
```suggestion
    `T` with `*std::move(other)`.
```

This one is stylistic, doesn't really matter. It means the same thing either way. `std::move(*other)` would be the same too.

DRAFT.md
```diff
@@ -806,7 +804,7 @@ constexpr indirect(allocator_arg_t, const Allocator& a,
 
 20. _Effects_: `alloc` is direct-non-list-initialized with `a`. If
     `other` is valueless, `*this` is valueless. Otherwise, copy constructs an
```
jwakely, line 806:
```suggestion
    `other` is valueless, `*this` is valueless. Otherwise, constructs an
```

DRAFT.md
```diff
@@ -784,8 +783,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a, in_place_t, Us&
 15. _Mandates_: `T` is a complete type.
 
 16. _Effects_: `alloc` is direct-non-list-initialized with `a`.
-    Direct-non-list-initializes an owned object of type `T` using the specified
-    allocator with `std​::​forward<Us>(us...)`.
+    Direct-non-list-initializes an owned object of type `T` with `std​::​forward<Us>(us...)`.
```
jwakely, line 786:
```suggestion
    Constructs an owned object of type `T` with `std​::​forward<Us>(us)...`.
```
N.B. the `...` needs to be _outside_ the parens

DRAFT.md
```diff
@@ -752,7 +751,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 7. _Mandates_: `T` is a complete type.
 
 8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Value
-   initializes an owned object of type `T` using the specified allocator.
+   initializes an owned object of type `T`.
```
jwakely, line 754: I think this should be "constructs" not "value-initializes" because the latter prevents the allocator's construct member from providing additional args. Value-initialization means precisely `T()` and nothing else. Maybe "Constructs an owned object of type T with an empty argument list" ?

```suggestion
   Constructs an owned object of type `T` with an empty argument list.
```

DRAFT.md
```diff
@@ -752,7 +751,7 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 7. _Mandates_: `T` is a complete type.
 
 8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Value
```
jwakely, line 753:
```suggestion
8. _Effects_: `alloc` is direct-non-list-initialized with `a`. 
```

Github won't let me make a suggestion that spans this line and the next green one, because of the red line between them, but I think this word should go as explained next...

DRAFT.md
```diff
@@ -1230,7 +1227,7 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a);
...
```
jwakely, line 1229:
```suggestion
8. _Effects_: `alloc` is direct-non-list-initialized with `a`.
```

DRAFT.md
```diff
@@ -1230,7 +1227,7 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a);
...
-initializes an owned object of type `T` using the specified allocator.
+initializes an owned object of type `T`.
```
jwakely, line 1230:
```suggestion
Constructs an owned object of type `T` with an empty argument list.
```

DRAFT.md
```diff
@@ -1263,8 +1260,7 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a,
 14. _Mandates_: `T` is a complete type.
 
 15. _Effects_: `alloc` is direct-non-list-initialized with `a`.
-    Direct-non-list-initializes an owned object of type `U` using the specified
-    allocator with `std​::​forward<Ts>(ts...)`.
+    Direct-non-list-initializes an owned object of type `U` with `std​::​forward<Ts>(ts...)`.
```
jwakely, line 1263:
```suggestion
    Constructs an owned object of type `U` with `std​::​forward<Ts>(ts)...`.
```

DRAFT_COROLLARY.md
```diff
@@ -173,8 +173,8 @@ D. _Constraints_: `is_constructible_v<T, U>` is true.
 E. _Mandates_: `T` is a complete type.
 
 F. _Effects_: `alloc` is direct-non-list-initialized with `a`.
-    Direct-non-list-initializes an owned object of type `T` using the specified
-    allocator with `std​::​forward<U>(u)`.
+    Direct-non-list-initializes an owned object of type `T` with
```
jwakely, line 176:
```suggestion
    Constructs an owned object of type `T` with
```

DRAFT_COROLLARY.md
```diff
@@ -205,8 +205,8 @@ J. _Constraints_: `is_copy_constructible_v<T>` is `true`.
 K. _Mandates_: `T` is a complete type.
 
 L. _Effects_: `alloc` is direct-non-list-initialized with `a`.
-    Direct-non-list-initializes an owned object of type `T` using the specified
-    allocator with `ilist, std​::​forward<U>(u)`.
+    Direct-non-list-initializes an owned object of type `T` with `ilist,
```
jwakely, line 208:
```suggestion
    Constructs an owned object of type `T` with `ilist,
```

jwakely (changes requested):
As discussed on mattermost, explicitly saying "value-initializes" and "direct-non-list-initializes with [args]" implies _exactly_ that form of construction, which prevents a scoped allocator from doing its thang in its `construct` member.

Twon (approved):
Read well to me know

Twon (dismissed):

Twon (approved):

DRAFT.md
```diff
@@ -623,15 +631,15 @@ move assignment, or swapping of the allocator only if
     is `true` within the implementation of the corresponding indirect value
     operation.
 
-4. A program that instantiates the definition of indirect for a non-object type,
+5. A program that instantiates the definition of indirect for a non-object type,
    an array type, or a cv-qualified type is ill-formed.
```
jwakely, line 635: We discussed adding this here, but maybe you want to do that separately:

```suggestion
   an array type, `in_place_t`, or a cv-qualified type is ill-formed.
```

DRAFT.md
```diff
@@ -751,8 +759,8 @@ explicit constexpr indirect(allocator_arg_t, const Allocator& a);
 
 7. _Mandates_: `T` is a complete type.
 
-8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Value
-   initializes an owned object of type `T` using the specified allocator.
+8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs owned
```
jwakely, line 762:
```suggestion
8. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs an owned
```

DRAFT.md
```diff
@@ -854,21 +862,20 @@ constexpr indirect& operator=(const indirect& other);
   If `std::allocator_traits<Alloc>::propagate_on_container_copy_assignment` is
   `true` and `alloc != other.alloc` then the allocator needs updating.
 
-  If `other` is valueless, `*this` becomes valueless and the owned value in this,
-  if any, is destroyed using `allocator_traits<allocator_type>::destroy` and then
-  deallocated using `allocator_traits<allocator_type>::deallocate`.
+  If `other` is valueless, `*this` becomes valueless and the owned value in
```
jwakely, line 865: Other places talk about "owned object" not "owned value".

DRAFT.md
```diff
@@ -854,21 +862,20 @@ constexpr indirect& operator=(const indirect& other);
...
+  this, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
+  and then deallocated using `allocator_traits<allocator_type>::deallocate`.
```
jwakely, line 867:
```suggestion
  and then the storage is deallocated using `allocator_traits<allocator_type>::deallocate`.
```

DRAFT.md
```diff
@@ -854,21 +862,20 @@ constexpr indirect& operator=(const indirect& other);
...
   `allocator_traits<allocator_type>::construct` with the owned object from
-  `other` as the argument, with memory allocated using either the allocator in
-  `this` or the allocator in `other` if the allocator needs updating. The
-  previously owned object in this, if any, is destroyed using
-  `allocator_traits<allocator_type>::destroy` and then deallocated using
-  `allocator_traits<allocator_type>::deallocate`.
+  `other` as the argument, using either the allocator in `this` or the allocator
+  in `other` if the allocator needs updating. The previously owned object in
+  this, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
+  and then deallocated using `allocator_traits<allocator_type>::deallocate`.
```
jwakely, line 878:
```suggestion
  and then the storage is deallocated using `allocator_traits<allocator_type>::deallocate`.
```

DRAFT.md
```diff
@@ -892,9 +899,9 @@ constexpr indirect& operator=(indirect&& other) noexcept(
   `std::allocator_traits<Alloc>::propagate_on_container_move_assignment` is
   `true` and `alloc != other.alloc` then the allocator needs updating.
 
-  If `other` is valueless, `*this` becomes valueless and the owned value in this,
-  if any, is destroyed using `allocator_traits<allocator_type>::destroy` and then
-  deallocated using `allocator_traits<allocator_type>::deallocate`.
+  If `other` is valueless, `*this` becomes valueless and the owned value in
+  this, if any, is destroyed using `allocator_traits<allocator_type>::destroy`
+  and then deallocated using `allocator_traits<allocator_type>::deallocate`.
```
jwakely, line 904:
```suggestion
  and then the storage is deallocated using `allocator_traits<allocator_type>::deallocate`.
```

Since you already said the allocator is used, maybe all these cases could just say "the storage is deallocated" and not mention how. There's no other way to do it except the allocator's `deallocate` member.

DRAFT.md
```diff
@@ -904,9 +911,9 @@ constexpr indirect& operator=(indirect&& other) noexcept(
   Otherwise , if `alloc != other.alloc` or `this` is valueless, a new owned
   object is constructed in `this` using
   `allocator_traits<allocator_type>::construct` with the owned object from
-  `other` as the argument as an rvalue, with memory allocated using either the
-  allocator in `this` or the allocator in `other` if the allocator needs
-  updating. The previous owned object in this, if any, is destroyed using
+  `other` as the argument as an rvalue, using either the allocator in `this` or
+  the allocator in `other` if the allocator needs updating. The previous owned
+  object in this, if any, is destroyed using
```
jwakely, line 916:
```suggestion
  object in `*this`, if any, is destroyed using
```

jwakely: I think most mentions of `this` should be `*this`, since the former is just a pointer and so doesn't own anything.

DRAFT.md
```diff
@@ -591,24 +599,24 @@ become valueless after it has been moved from.
 `allocator_traits<Allocator>::value_type` is not the same type as `T`, the
 program is ill-formed. Every object of type `indirect<T, Allocator>` uses an
 object of type `Allocator` to allocate and free storage for the owned object as
-needed. The owned object is constructed using the function
-`allocator_traits<allocator_type>::construct` and destroyed using the function
-`allocator_traits<allocator_type>::destroy`.
+needed.
 
-// DRAFTING NOTE: [indirect.general]#3 modeled on [container.reqmts]#64
+3. Constructing an owned object using an allocator `a` and arguments `args...`
```
jwakely, line 604: All the uses in the constructors below say "with arguments ..., using the allocator `alloc`" but this has the two terms the other way around. I think the way you used it later is better, so should change this here:

```suggestion
3. Constructing an owned object with arguments `args...` using an allocator `a` 
```

DRAFT.md
```diff
@@ -1104,19 +1111,19 @@ object may only become valueless after it has been moved from.
...
-`allocator_traits<allocator_type>::rebind_traits<U>::construct` and destroyed
- using the function
-`allocator_traits<allocator_type>::rebind_traits<U>::destroy`, where `U` is
-either `allocator_type::value_type` or an internal type used by the polymorphic
-value.
-
-3. Copy constructors for a polymorphic value obtain an allocator by calling
+needed.
+
+3. Constructing an owned object using an allocator `a` and arguments `args...`
```
jwakely, line 1116:
```suggestion
3. Constructing an owned object with arguments `args...` using an allocator `a`
```

DRAFT.md
jwakely: Oh, especially as it should really be `allocator_traits<allocator_type>::rebind_traits<T>::deallocate`  for `polymorphic`, which is not correct below!

So not mentioning `deallocate` by name avoids that problem.

DRAFT.md
```diff
@@ -1380,9 +1388,9 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
   Otherwise if `alloc != other.alloc`; if `other` is not valueless, a new owned
   object is constructed in `this` using
   `allocator_traits<allocator_type>::construct` with the owned object from
-  `other` as the argument as an rvalue, with memory allocated using either the
-  allocator in `this` or the allocator in `other` if the allocator needs
-  updating. The previous owned object in this, if any, is destroyed using
+  `other` as the argument as an rvalue, using either the allocator in `this` or
+  the allocator in `other` if the allocator needs updating. The previous owned
+  object in this, if any, is destroyed using
```
jwakely, line 1393:
```suggestion
  object in `*this`, if any, is destroyed using
```

And change other cases of `this` to `*this` in this paragraph.

DRAFT.md
```diff
@@ -1301,14 +1309,14 @@ constexpr polymorphic(allocator_arg_t, const Allocator& a,
...
-    `other` is valueless, `*this` is valueless. Otherwise, if `alloc ==
-    other.alloc` either constructs an object of type `polymorphic` that owns the
-    owned value of other, making `other` valueless; or, owns an object of the
-    same type constructed from the owned value of `other` using the specified
-    allocator, considering that owned value as an rvalue. Otherwise if `alloc !=
-    other.alloc`, constructs an object of type `polymorphic` using the specified
-    allocator, considering that owned value as an rvalue.
+21. _Effects_: `alloc` is direct-non-list-initialized with `a`. If `other` is
+    valueless, `*this` is valueless. Otherwise, if `alloc == other.alloc` either
+    constructs an object of type `polymorphic` that owns the owned value of
```
jwakely, line 1314: "owned value" vs "owned object"

DRAFT.md
```diff
@@ -1380,9 +1388,9 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
...
   object is constructed in `this` using
   `allocator_traits<allocator_type>::construct` with the owned object from
-  `other` as the argument as an rvalue, with memory allocated using either the
-  allocator in `this` or the allocator in `other` if the allocator needs
-  updating. The previous owned object in this, if any, is destroyed using
+  `other` as the argument as an rvalue, using either the allocator in `this` or
+  the allocator in `other` if the allocator needs updating. The previous owned
+  object in this, if any, is destroyed using
   `allocator_traits<allocator_type>::destroy` and then deallocated using
   `allocator_traits<allocator_type>::deallocate`.
```
jwakely, line 1395: As noted above, this should be  `allocator_traits<allocator_type>::rebind_traits::deallocate` but maybe just say "the storage is deallocated" instead.

jwakely (changes requested):

DRAFT.md
jbcoe: Now use "owned object" throughout

nbx8 (approved):
Wording looks fine and consistent.

DRAFT.md
```diff
@@ -1104,19 +1111,19 @@ object may only become valueless after it has been moved from.
...
- using the function
-`allocator_traits<allocator_type>::rebind_traits<U>::destroy`, where `U` is
-either `allocator_type::value_type` or an internal type used by the polymorphic
-value.
-
-3. Copy constructors for a polymorphic value obtain an allocator by calling
+needed.
+
+3. Constructing an owned object with arguments `args...` using an allocator `a`
+means calling `allocator_traits<allocator_type>::rebind_traits<U>::construct(a,
```
jwakely, line 1117: Strictly speaking, you don't need to rebind the traits here. The `construct` member is a function template so that `Alloc<T>` can construct+destroy objects of type `U`. So it's up to you whether you want to include the rebind for consistency/simplicity, or omit it (which would mean slightly adjusting the "and U is either ..." text to only refer to the `allocate` call). The `allocate` call below this does need to use the rebound type (and it does so that's good).

DRAFT.md
```diff
@@ -1375,17 +1383,17 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
...
   If `alloc == other.alloc`, swaps the owned objects in `this` and `other`; the
   owned object in `other`, if any, is then destroyed using
-  `allocator_traits<allocator_type>::destroy` and then deallocated using
-  `allocator_traits<allocator_type>::deallocate`.
+  `allocator_traits<allocator_type>::rebind_traits::destroy` and then the
+  storage is deallocated.
 
-  Otherwise if `alloc != other.alloc`; if `other` is not valueless, a new owned
+  Otherwise, if `alloc != other.alloc`; if `other` is not valueless, a new owned
   object is constructed in `this` using
```
jwakely, line 1390:
```suggestion
  object is constructed in `*this` using
```

DRAFT.md
```diff
@@ -1375,17 +1383,17 @@ constexpr polymorphic& operator=(polymorphic&& other) noexcept(
...
-  `allocator_traits<allocator_type>::construct` with the owned object from
-  `other` as the argument as an rvalue, with memory allocated using either the
-  allocator in `this` or the allocator in `other` if the allocator needs
-  updating. The previous owned object in this, if any, is destroyed using
-  `allocator_traits<allocator_type>::destroy` and then deallocated using
-  `allocator_traits<allocator_type>::deallocate`.
+  `allocator_traits<allocator_type>::rebind_traits::construct` with the owned
+  object from `other` as the argument as an rvalue, using either the allocator
+  in `this` or the allocator in `other` if the allocator needs updating. The
```
jwakely, line 1393:
```suggestion
  in `*this` or the allocator in `other` if the allocator needs updating. The
```

jwakely (changes requested):

DRAFT_COROLLARY.md
```diff
@@ -277,7 +279,6 @@ explicit constexpr polymorphic(allocator_arg_t, const Allocator& a, U&& u);
 
 D. _Constraints_: `is_base_of_v<T, std::remove_cvref_t<U>>` is `true`.
    `is_copy_constructible_v<remove_cvref_t<U>>` is `true`.
-   `is_same_v<remove_cvref_t<U>, in_place_t>` is `false`.
```
jwakely, line 280: Should this have `remove_cvref_t<U>` is not a specialization of `in_place_type_t`? And similarly for the ctor above.

DRAFT_COROLLARY.md
```diff
@@ -213,11 +214,12 @@ template <class U>
 constexpr indirect& operator=(U&& u);
 ```
 
-A. _Constraints_: `is_constructible<T, U>` is true.
-  `is_assignable<T,U>` is true.
+A. _Constraints_: `is_constructible<T, U>` is true. `is_assignable<T,U>` is
```
jwakely, line 217:
```suggestion
A. _Constraints_: `is_constructible_v<T, U>` is true. `is_assignable_v<T,U>` is
```

DRAFT.md
```diff
@@ -820,11 +831,11 @@ constexpr indirect(allocator_arg_t, const Allocator& a, indirect&& other)
   noexcept(allocator_traits<Allocator>::is_always_equal);
```
jwakely, line 831:
```suggestion
  noexcept(allocator_traits<Allocator>::is_always_equal::value);
```

DRAFT_COROLLARY.md
```diff
@@ -150,14 +151,14 @@ explicit constexpr indirect(U&& u);
 ```
 
 A. _Constraints_: `is_constructible_v<T, U>` is true.
-   `is_copy_constructible_v<U>` is `true`.
-   `is_same_v<remove_cvref_t<U>, in_place_t>` is `false`.
-   `is_same_v<remove_cvref_t<U>, indirect>` is `false`.
+   `is_copy_constructible_v<U>` is `true`. `is_same_v<remove_cvref_t<U>,
```
jwakely, line 154:
```suggestion
   `is_copy_constructible_v<T>` is `true`. `is_same_v<remove_cvref_t<U>,
```

DRAFT_COROLLARY.md
```diff
@@ -171,18 +172,18 @@ D. _Constraints_: `is_constructible_v<T, U>` is true.
...
-F. _Effects_: `alloc` is direct-non-list-initialized with `a`.
-    Direct-non-list-initializes an owned object of type `T` using the specified
-    allocator with `std​::​forward<U>(u)`.
+F. _Effects_: `alloc` is direct-non-list-initialized with `a`. Constructs an
+    owned object of type `T` with `std​::​forward<U>(u)`, using the allocator
+    `alloc`.
 
 ```c++
 template<class I, class... Us>
 explicit constexpr indirect(in_place_t, std::initializer_list<I> ilist,
```
jwakely, line 181:
```suggestion
explicit constexpr indirect(in_place_t, initializer_list<I> ilist,
```

jwakely: and line 197 too

DRAFT.md
```diff
@@ -892,26 +902,23 @@ constexpr indirect& operator=(indirect&& other) noexcept(
...
 
-  Otherwise if `alloc == other.alloc`, swaps the owned objects in `this` and
+  Otherwise, if `alloc == other.alloc`, swaps the owned objects in `this` and
   `other`; the owned object in `other`, if any, is then destroyed using
-  `allocator_traits<allocator_type>::destroy` and then deallocated using
-  `allocator_traits<allocator_type>::deallocate`.
+  `allocator_traits<allocator_type>::destroy` and then the storage is deallocated.
 
-  Otherwise , if `alloc != other.alloc` or `this` is valueless, a new owned
+  Otherwise, if `alloc != other.alloc` or `this` is valueless, a new owned
```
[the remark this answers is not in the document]
jwakely: Actually, is the "or `*this` is valueless" condition unreachable? We have:

if (other is valueless) { }
else if (alloc == other.alloc) { }
else if (alloc != other.alloc || *this is valueless) { }

But we know by the time we get to the third `if` that alloc != other.alloc because we didn't take the second branch, so it can be simply:

if (other is valueless) { }
else if (alloc == other.alloc) { }
else { }

So the line above can be simple "Otherwise, a new owned ..."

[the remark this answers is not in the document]
jwakely: I think the same observation is true for the copy assignment operator, but at least there the "alloc != other.alloc or *this is not valueless" is the inverse of the previous condition. Here on the move assignment operator it looks like a copy & paste error because "or *this is valueless" makes no sense here.

DRAFT_COROLLARY.md
```diff
@@ -213,11 +214,12 @@ template <class U>
...
 
-A. _Constraints_: `is_constructible<T, U>` is true.
-  `is_assignable<T,U>` is true.
+A. _Constraints_: `is_constructible<T, U>` is true. `is_assignable<T,U>` is
+  true.
 
 B. _Mandates_: `T` is a complete type.
 
+
 C. _Effects_: If `this` is valueless then equivalent to
```
jwakely, line 223:
```suggestion
C. _Effects_: If `*this` is valueless then equivalent to
```

DRAFT_COROLLARY.md
```diff
@@ -213,11 +214,12 @@ template <class U>
...
    `*this = indirect<T>(std::forward<U>(u))`.
```
jwakely, line 224: This uses a default constructed allocator, which is not listed as a constraint of calling this assignment operator. Should it be `*this = indirect(allocator_arg, alloc, std::forward<U>(u));` or should it be constrained to require `is_default_constructible_v<Allocator>`? Or something else?

DRAFT.md
```diff
@@ -983,8 +990,8 @@ constexpr void swap(indirect& other) noexcept(
   || allocator_traits::is_always_equal::value);
```
jwakely, line 990:
```suggestion
  || allocator_traits<Allocator>::is_always_equal::value);
```

And on the line above, which I can't suggest a change for because Github thinks it's not part of the review.

DRAFT.md
```diff
@@ -1302,14 +1310,14 @@ constexpr polymorphic(allocator_arg_t, const Allocator& a,
                       polymorphic&& other) noexcept(allocator_traits::is_always_equal::value);
```
jwakely, line 1310:
```suggestion
                      polymorphic&& other) noexcept(allocator_traits<Allocator>::is_always_equal::value);
```

DRAFT_COROLLARY.md
jbcoe: `is_base_of_v<T, std::remove_cvref_t<U>>` is `true` should cover us as `in_place_type_t` is not derived from `T` or equal to `T` for any legal choice of `T`.

We can add the constraint if it helps to be explicit.

DRAFT_COROLLARY.md
jbcoe: Applied manually.

DRAFT.md
jbcoe: Applied manually.

DRAFT_COROLLARY.md
jbcoe: Nicely spotted. Agreed and applied manually.

DRAFT_COROLLARY.md
jbcoe: Agree. `is_base_of_v<T, std::remove_cvref_t<U>>` is `true` should have use covered but no harm in being explicit - better errors that way.

DRAFT.md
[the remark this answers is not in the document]
jbcoe: Agreed and fixed.

nbx8 (approved):
Wording changes look fine.

Twon (approved):
As far as I can see the changes look good.  Seems to be quite a bit of noise created by trivial formatting changes (line wrapping text), it would be useful in future to separate these out from actual changes to text.

jbcoe:
> As far as I can see the changes look good. Seems to be quite a bit of noise created by trivial formatting changes (line wrapping text), it would be useful in future to separate these out from actual changes to text.

Agreed. This PR got too big too fast. Thanks for review.

PR #399, opened 2024-03-20
Implement code review feedback from J.Wakely

Twon:


DRAFT_COROLLARY.md
```diff
@@ -157,17 +157,15 @@ A. _Constraints_: `is_constructible_v<T, U>` is true.
...
 ```c++
 template <class U = T>
-explicit constexpr indirect(allocator_arg_t, const Allocator& a,
-                            in_place_t, U&& u);
+explicit constexpr indirect(allocator_arg_t, const Allocator& a, U&& u);
 ```
 
 D. _Constraints_: `is_constructible_v<T, U>` is true.
-   `is_copy_constructible_v<U>` is `true`.
-   `is_same_v<remove_cvref_t<U>, in_place_t>` is `false`.
```
jwakely, line 170: Was it intentional to remove line 170?

DRAFT_COROLLARY.md
```diff
@@ -269,8 +266,8 @@ A. _Constraints_: `is_base_of_v<T, std::remove_cvref_t<U>>` is `true`.
 
 B. _Mandates_: `T` is a complete type.
 
-C. _Effects_: Equivalent to `polymorphic(std::allocator_arg_t{}, A{},
-   std::in_place_type_t<std::remove_cvref_t<U>>{}, std::forward<U>(u))`.
+C. _Effects_: Equivalent to `polymorphic(allocator_arg, A{},
+   in_place_type<std::remove_cvref_t<U>>, std::forward<U>(u))`.
```
jwakely, line 270: Remove std:: on remove_cvref_t (keep it on std::forward, because of LWG policy that nobody really likes any more!)

DRAFT_COROLLARY.md
jwakely: I think you still want that constraint, so that this constructor is not used for:
`indirect<T> i(allocator_arg, alloc, std::in_place);`
because that should go to the `indirect(allocator_arg_t, const Allocator& a, in_place_t, Us&& ...us);` that's already in P3019.

Twon: Yes, because as your code review pointed out this parameter was missing from the synopsis: https://github.com/jbcoe/value_types/commit/3474b5950948fd78c4168904c27d81e52b3c5564#r140002552

I have also removed this from the function arguments because this matches the definition for polymorphic now.

DRAFT_COROLLARY.md
Twon: I will raise a new PR to address this.  Thanks for catching, I missed this.

DRAFT_COROLLARY.md
Twon: Sorry, your second comment was not visible when I answered initially.  But I think it is correct to remove this so that `indirect` and `polymorphic` versions of these functions are aligned.  That alternative is to add the `in_place_t` parameter in the synopsis and then also to `polymorphic` equivalent overload and synopsis.  

Does is make more sense to keep the `in_place_t` (and thus the constraint here)?

jbcoe: I agree with JW that
`indirect<T> i(allocator_arg, alloc, std::in_place);` should go to `indirect(std::allocator_arg_t, const Allocator& a, in_place_t, Us&& ...us);`

The same should be the case for `polymorphic` so we need the constraint in both places.

jwakely: No, the `in_place_t` parameter should not be there, but you still need to check that U is not that type. See https://eel.is/c++draft/optional.optional#optional.ctor-23

jwakely: It's precisely because the `in_place_t` parameter is _not_ there that you need the constraint.

jbcoe: We probably need to make `indirect<in_place_t>` illegal too.

Twon: Thanks for the clarification.  Constraint reinstated and added to the equivalent constructor in `polymorphic`: https://github.com/jbcoe/value_types/pull/400

Twon: > We probably need to make `indirect<in_place_t>` illegal too.

I think that is covered by the constraint: `is_same_v<remove_cvref_t<U>, indirect>` is `false`

Twon: Thanks, I appreciate the guidance here.  I have now added these to the general section for both types as you have outlined.

PR #402, opened 2024-03-20
Minor fixes

jwakely:
The standard uses `class` in template heads, not `typename`. And `std::` should not be present on `std::allocator_arg_t` or `std::allocator_traits`.


jbcoe (approved):
Thanks for these fixes.

PR #426, opened 2024-03-23
Liberally apply [[nodiscard]]

jbcoe:


Twon:
Why does this not cause an errors for C++14 builds given `[[nodiscard]]` is a C++17 attribute: https://en.cppreference.com/w/cpp/language/attributes/nodiscard ?

jbcoe:
> Why does this not cause an errors for C++14 builds given `[[nodiscard]]` is a C++17 attribute: https://en.cppreference.com/w/cpp/language/attributes/nodiscard ?

Because attributes can be ignored.

Twon:
> > Why does this not cause an errors for C++14 builds given `[[nodiscard]]` is a C++17 attribute: https://en.cppreference.com/w/cpp/language/attributes/nodiscard ?
> 
> Because attributes can be ignored.

Wondered if that was the case, thanks for confirming...

Twon (approved):

PR #435, opened 2024-09-28
Add build scripts for ease of use

jbcoe:


nbx8 (approved):
Build scripts seem fine.

PR #483, opened 2024-12-15
Fix bazel deps

jbcoe:
Fix bazel deps following https://github.com/google/googletest/issues/3824


jbcoe:
Odd code coverage errors in build will be fixed in a follow-up PR. They are not bazel-related.

PR #486, opened 2025-01-01
Update coverage to use fastcov and avoid external coverage library

Twon:


.github/workflows/code_coverage.yml
```diff
@@ -56,15 +56,23 @@ jobs:
...
+        run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target coverage-lcov
+
+      - name: Cobertura
+        # Run Cobertura target to generate code coverage report in XML
+        run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target coverage-cobertura
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4.4.1
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
Twon, line 74: @jbcoe This will need retrieving from the codecov website and see in the CI secrets

jbcoe:
We have 

```
ModuleNotFoundError: No module named 'distutils'
```

`distutils` was removed in Python 3.12 - are we using too new a Python version?

.github/workflows/code_coverage.yml
jbcoe: I added `CODECOV_TOKEN` to repository secrets.

Twon:
> We have
> 
> ```
> ModuleNotFoundError: No module named 'distutils'
> ```
> 
> `distutils` was removed in Python 3.12 - are we using too new a Python version?

I've added this to the requirements now and that has resolved the issue.

.github/workflows/code_coverage.yml
Twon: Now this needs setting to resolve the final error:

```
error - 2025-01-01 19:56:18,903 -- Upload failed: {"message":"Token required because branch is protected"}
```

Twon:
Looks like the upload is now successful:

```
info - 2025-01-01 20:06:17,076 -- Your upload is now processing. When finished, results will be available at: https://app.codecov.io/github/jbcoe/value_types/commit/e099ddb8e8d241e4319a9807444134d7c8802387
```

jbcoe:
> Looks like the upload is now successful:
> 
> ```
> info - 2025-01-01 20:06:17,076 -- Your upload is now processing. When finished, results will be available at: https://app.codecov.io/github/jbcoe/value_types/commit/e099ddb8e8d241e4319a9807444134d7c8802387
> ```

I'm seeing some issues with the uploaded report: https://app.codecov.io/github/jbcoe/value_types/commit/e099ddb8e8d241e4319a9807444134d7c8802387

Not sure what's going on here.

jbcoe:
I think we need to opt in to coverage. We're currently reporting low coverage as we're including benchmarks and the like. Previously we had files opted in to coverage when used in tests, can we do something similar again.

Awesome work getting this set up in any case - looks much easier to use.

Twon:
> I think we need to opt in to coverage. We're currently reporting low coverage as we're including benchmarks and the like. Previously we had files opted in to coverage when used in tests, can we do something similar again.
> 
> Awesome work getting this set up in any case - looks much easier to use.

Ah, it makes sense, as I just search for all targets and add them to coverage. This works for other projects where I use this as dependencies come via Conan. But as we use FetchContent for this project these projects become targets to this project. 

I think the answer is as you say to opt it. This just requires moving the code which adds coverage setting to all targets out to a helper function which we call for targets like the old coverage library. I can take a look at this shortly

jbcoe:
Opt-in implmeneted in https://github.com/jbcoe/value_types/pull/487

jbcoe:
Looks like maybe the uploaded report is not in the expected format (XML vs JSON?). I'm not sure what might be done.

Twon:
> Looks like maybe the uploaded report is not in the expected format (XML vs JSON?). I'm not sure what might be done.

Cobetura (XML format) is supported by Codecov. I'm not sure what the unsupported format issue is as I have this working on other repositories. We can also output to other formats and see if they work. We can try uploading the raw lcov file

jbcoe:
Fixed by just using the XML.

jbcoe (approved):
Awesome work - thanks Ant.

PR #487, opened 2025-01-01
Reinstate registering of targets

jbcoe:


cmake/coverage.cmake
```diff
@@ -338,6 +338,15 @@ function(coverage_target_clean_intermediate_file)
 endfunction()
 
 
+set_property(GLOBAL PROPERTY global_coverage_test_targets)
```
jbcoe, line 341: I'm not sure if there's a neater way of doing this.

Twon (approved):
Thanks for this, looks like exactly what is needed

cmake/coverage.cmake
Twon: > I'm not sure if there's a neater way of doing this.

I think this is the least invasive approach.

PR #478, opened 2024-12-01
[do not merge] Remove deferred requirements, use static_assert instead

jbcoe:
I think this is the direction we might want to pursue. It means that traits might lie and that indirect can be used with non-copyable `T`.

Innovation around requirements/constraints might be better deferred to another proposal where other standard library classes can modified in sync.

Associated issues: 
* #479 
* #471 
* #482

PR #489, opened 2025-01-02
Fix indirect comparisons

jbcoe:
Do not merge until https://github.com/jbcoe/value_types/pull/478 is submitted.


nbx8 (approved):
All looks good.

DRAFT.md
```diff
@@ -867,12 +869,12 @@ class indirect {
   template <class U, class AA>
   friend constexpr auto operator<=>(
     const indirect& lhs, const indirect<U, AA>& rhs) noexcept(see below)
-    -> compare_three_way_result_t<T, U>;
+    -> synth-three-way-result<T, U>;
```
Twon, line 872: I thought we had previously made this change in the LWG session in Tokyo.  This is want we want.

Twon (commented):
Removing the constraints here seems fine to me.  Given we return `synth-three-way-result` I'm not sure what we get by additionally specifying it as convertible to bool.

Twon (commented):
Looks good

Twon (approved):

PR #490, opened 2025-01-12
Amend draft to add missing explicit on indirect ctor in synopsis

jbcoe:
Use "may be X only" rather that "may only be X".


jbcoe:
Approved by @Twon indirectly in #478

PR #492, opened 2025-01-12
Restore legacy C++ implementations

jbcoe:


Twon (approved):
Good to see these types reinstated

Issue #503, opened 2025-01-18
Clarify exception specification for `operator<=>` with `synth-three-way`


Ukilele:
Hey, this is an issue about the wording. It's about the noexcept specification for the two `operator<=>`'s.
In [X.Y.8 Relational operators [indirect.relops]](https://github.com/jbcoe/value_types/blob/main/DRAFT.md#xy8-relational-operators-indirectrelops) the noexcept specification of `operator<=>` is defined as: `noexcept(noexcept(synth-three-way(*lhs, *rhs)));`. And in [X.Y.9 Comparison with T [indirect.comp.with.t]](https://github.com/jbcoe/value_types/blob/main/DRAFT.md#xy9-comparison-with-t-indirectcompwitht) it is defined as `noexcept(noexcept(synth-three-way(*lhs, rhs)));`.

I looked up the definition of `synth-three-way` at https://eel.is/c++draft/expos.only.entity. And it is a lambda without any noexcept specification. Which means that the noexcept specification of it's callable operator is implicitly `noexcept(false)`. Which means that `noexcept(synth-three-way(*lhs, *rhs))` will always yield `false`. I assume that this is not intended. Here are a few other specifications of `operator<=>` from the STL, which use `synth-three-way`: [reference_wrapper](https://eel.is/c++draft/full#refwrap.comparisons-itemdecl:4), [tuple](https://eel.is/c++draft/full#lib:operator%3c=%3e,tuple), [pair](https://eel.is/c++draft/pairs#lib:operator%3c=%3e,pair), [array](https://eel.is/c++draft/array.syn).

I am not sure, but my guess would be, to drop the noexcept specification from the two `operator<=>`'s. At least that would be consistent with the existing STL.

Blocks [Implement operator <=> in indirect to use synth-three-way-result #498](https://github.com/jbcoe/value_types/issues/498.)

jbcoe:
Thanks Kilian, this sounds like the right solution.

Are you able to propose a PR for the draft?

PR #504, opened 2025-01-18
Remove noexcept specification from operator<=>

Ukilele:
Note that I did not update the revision history. I thought that following bullet point might cover my change:
```markdown
* Fix specification of `<=>` to use `synth-three-way-result`.
```
I you think otherwise, I am happy to add a new bullet point.

Fixes https://github.com/jbcoe/value_types/issues/503

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -860,12 +860,12 @@ class indirect {
 
   template <class U, class AA>
   friend constexpr auto operator<=>(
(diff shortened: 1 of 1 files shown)


jbcoe:
Thanks Kilian, we’ll need a new bullet point in the revision history as this will bump the revision number to r13.

jbcoe (approved):
Great work @Ukilele thanks so much for spotting and addressing this.

PR #507, opened 2025-01-23
Fix typo `alloc` -> `a`

Ukilele:
Just a very small typo fix. Please let me know if you need a new bullet point for this in the revision history `Changes in R13`. But I assume not.


jbcoe:
Nicely spotted @Ukilele 

As you say, no need for changelog entries here, it just becomes noise.

PR #510, opened 2025-01-28
Sort constraints, remove needless constraints

### LINKED ISSUE #509: Constraint "`is_same_v<U, polymorphic>` is `false`" only available on two of the constructors taking `in_place_type_t<U>`
`polymorphic` has four constructors which take a `std::in_place_type_t<U>` as function parameter. They differ by the fact whether they additionally take an allocator and an initializer_list or not.

```c++
  template <class U, class... Ts>
  explicit constexpr polymorphic(in_place_type_t<U>, Ts&&... ts);

  template <class U, class... Ts>
  explicit constexpr polymorphic(allocator_arg_t, const Allocator& a,
                                 in_place_type_t<U>, Ts&&... ts);

  template <class U, class I, class... Us>
  explicit constexpr polymorphic(in_place_type_t<U>,
                                 initializer_list<I> ilist, Us&&... us);

  template <class U, class I, class... Us>
  explicit constexpr polymorphic(allocator_arg_t, const Allocator& a,
                                 in_place_type_t<U>,
                                 initializer_list<I> ilist, Us&&... us);
```

But only the ones which take an `initializer_list<I>` have the constraint: `is_same_v<U, polymorphic> is false,`. I just wanted to double check, that this is intentional and that they are not missing on the other two constructors. And I am wondering what the purpose of this constraint it.

jbcoe:
fixes #509

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -49,6 +49,13 @@ should not be considered in isolation.
 
 * Add a second clarifying example to tagged constructors explanatory text.
 
+* Remove constraint `is_same_v<U, polymorphic>` is `false` on polymorphic constructors
+  taking `in_place_type_t` and an intializer list.
+
+* Order `polymorphic` constructor constraints consistently.
+
+* Remove needless introduction of `UU` in `indirect` constraints.
+
 ## Changes in R12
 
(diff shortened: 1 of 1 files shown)


Twon (approved):
Looks good

PR #516, opened 2025-02-06
Fix CI environment to avoid FetchPackage errors

jbcoe:
Avoid hitting errors on Windows CI builds:

```
SSL connect error. If this is due to https certificate verification failure, one may set environment variable CMAKE_TLS_VERIFY=0 to suppress it.
```


jbcoe:
Does not help.

PR #520, opened 2025-02-07
Exploration of deferred constraints and composition

jbcoe:
Add some of the test code we wrote when investigating deferred constraints


Twon (approved):
Nice to have a minimal reproduction case to highlight the issues.

PR #524, opened 2025-02-09
Notes for presentation

jbcoe:


Twon (approved):
This looks good to me, covers all of the salient points and explicit addresses the differences since R11.

PR #534, opened 2025-02-11
Update implementation of operator <=> to support types that do not have operator <=>

### LINKED ISSUE #498: Implement operator <=> in indirect to use synth-three-way-result
We can follow example code in https://en.cppreference.com/w/cpp/standard_library/synth-three-way and update tests to show that `<=>` can work for `indirect<T>` even when `T` does not support `<=>`.

jbcoe:
fixes #498 

CoAuthored with @Ukilele and @nbx8

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -44,6 +44,7 @@ namespace xyz {
 }
 #endif  // XYZ_UNREACHABLE_DEFINED
 
+namespace detail {
 template <class T, class U>
 concept three_way_comparable_with = requires(T& t, U& u) {
   { t <=> u } -> std::convertible_to<std::weak_ordering>;
@@ -68,6 +69,7 @@ constexpr auto synth_three_way(const T& t, const U& u)
 template <class T, class U = T>
 using synth_three_way_result =
     decltype(synth_three_way(std::declval<T&>(), std::declval<U&>()));
+}  // namespace detail
 
 template <class T, class A>
 class indirect;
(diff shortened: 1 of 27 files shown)


indirect.h
```diff
@@ -45,6 +45,34 @@ namespace xyz {
 }
 #endif  // XYZ_UNREACHABLE_DEFINED
 
+namespace detail {
+template <class T, class U>
+concept three_way_comparable_with = requires(T& t, U& u) {
```
Twon, line 50: What is the benefit of defining our own versions of these over using `std:: three_way_comparable`: https://en.cppreference.com/w/cpp/utility/compare/three_way_comparable

Also see, the three_way_comparable_with here matches `std:: three_way_comparable` and not `std:: three_way_comparable_with`

indirect.h
```diff
@@ -45,6 +45,34 @@ namespace xyz {
...
+constexpr auto synth_three_way(const T& t, const U& u)
+  requires requires {
+    { t < u } -> std::convertible_to<bool>;
+    { u < t } -> std::convertible_to<bool>;
+  }
+{
+  if constexpr (three_way_comparable_with<T, U>) {
+    return t <=> u;
+  } else {
+    if (t < u) return std::weak_ordering::less;
```
Twon, line 65: I think the else branch can just be `std:: compare_strong_order_fallback`: https://en.cppreference.com/w/cpp/utility/compare/compare_strong_order_fallback

jbcoe: std::compare_weak_order_fallback requires that they are the same type and we need to support comparison between different but comparable types.

Twon: Good point, thanks for the clarifying.

indirect.h
jbcoe: I can't get `std:: three_way_comparable` to play nice. Issues with conversion of types and recursion I think.

I'd like a working implementation merged before plenary: can we follow up and improve code when you're with us? It will be easier to discuss in person.

Twon: Yes, of course, only minor comments and not worth delaying for.

Twon (approved):

indirect.h
jbcoe: It would be nice to sort out but we can merge for now.

PR #541, opened 2025-02-12
Use std::three_way_comparable concept, do not define our own

jbcoe:
Constrain `indirect`'s `operator <=>` using SFINAE not concepts to avoid premature return-type evaluation.


Twon (approved):
Nice

indirect.h
```diff
@@ -410,12 +407,10 @@ class indirect {
...
   }
 
-  template <class U>
+  template <class U, typename std::enable_if<!is_indirect_v<U>, int>::type = 0>
   [[nodiscard]] friend constexpr auto operator<=>(const indirect<T, A>& lhs,
                                                   const U& rhs)
-      -> detail::synth_three_way_result<T, U>
-    requires(!is_indirect_v<U>)
-  {
+      -> detail::synth_three_way_result<T, U> {
```
ned14, line 413: First impression I got is why aren't your SFINAEing on the return type only?

You can then use concepts to control the return type availability, and this function will fail to substitute the return type, thus quietly suppressing the availability of this overload.

jbcoe: Do you have example code?

jbcoe: I think I've updated code to SFINAE on the return type only.

Maybe you can help me understand why that's better - I suppose it's a compile-time improving thing?

ned14: It's more a case of easiest than anything else.

Contracts early fail, SFINAE late fail.

SFINAE is therefore less noisy about failure.

If you want late silent failure, you put SFINAE outermost and put your noisy early stuff innermost.

It's also easier for debugging when the wrong overload get selected because you can expose the compiler diagnostics temporarily, fix what's wrong, re-silence the operation.

PR #544, opened 2025-02-12
Only run CI checks on relevant changes

### LINKED ISSUE #543: Avoid running expensive CI checks when they are not needed
We should set up CI to only run tests on affected targets: we don't need code coverage runs when a README is updated.

Currently this is done with exceptions, instead we need opt-in - only run tests if they are needed, not by default.

jbcoe:
Fixes #543

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -6,21 +6,21 @@ on:
   push:
     branches: [main]
     paths: |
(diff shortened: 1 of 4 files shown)


jbcoe:
We need a nice way of running checks defined by a `.yml` file if the `.yml` file is edited.

Twon (approved):
Looks sensible

PR #547, opened 2025-02-13
Generate and use compile_commands.json with CMake

jbcoe:
Generate compile_commands.json in CMake builds. 

Configure VsCode to read compile_commands.json from default CMake build directory.

### DIFF
--- a/scripts/cmake.sh
+++ b/scripts/cmake.sh
@@ -6,7 +6,7 @@ set -eux -o pipefail
 mkdir -p build
 
 # Generate build system specified in build directory with cmake
-cmake -Bbuild -GNinja
+cmake -Bbuild -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=1
 
 # Build the underlying build system via CMake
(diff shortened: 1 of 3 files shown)


scripts/cmake.sh
```diff
@@ -6,7 +6,7 @@ set -eux -o pipefail
 mkdir -p build
 
 # Generate build system specified in build directory with cmake
-cmake -Bbuild -GNinja
+cmake -Bbuild -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=1
```
Twon, line 9: We could use a CMakePresets.json to ensure this is always provided (we could also hide Ninja as the generator here, though I'd favour the "Ninja Muliti-Config" generator for only one build folder)

jbcoe: Feel free to amend the PR. Sounds like a neat approach.

jbcoe: We can improve this later. Merging now as I want this locally.

PR #548, opened 2025-02-13
Fix CMake warnings

jbcoe:
Fix warning for google test discovery following https://stackoverflow.com/questions/73457020/how-to-set-working-folder-when-running-ctest

Previously we saw:

```
CMake Warning (dev) at cmake/coverage.cmake:344:
  Syntax Warning in cmake code at column 50

  Argument not separated from preceding token by whitespace.
Call Stack (most recent call first):
  CMakeLists.txt:25 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

...

```
CMake Warning (dev) at /opt/homebrew/share/cmake/Modules/GoogleTest.cmake:559 (cmake_parse_arguments):
  The WORKING_DIRECTORY keyword was followed by an empty string or no value
  at all.  Policy CMP0174 is not set, so cmake_parse_arguments() will unset
  the arg_WORKING_DIRECTORY variable rather than setting it to an empty
  string.
Call Stack (most recent call first):
  cmake/xyz_add_test.cmake:103 (gtest_discover_tests)
  exploration/CMakeLists.txt:3 (xyz_add_test)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### DIFF
--- a/cmake/xyz_add_test.cmake
+++ b/cmake/xyz_add_test.cmake
@@ -101,7 +101,7 @@ function(xyz_add_test)
 
         include(GoogleTest)
         gtest_discover_tests(${XYZ_NAME}
-            WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+            WORKING_DIRECTORY $<TARGET_FILE_DIR:${XYZ_NAME}>
         )
 
     endif()
--- a/cmake/xyz_add_test.cmake
+++ b/cmake/xyz_add_test.cmake
@@ -88,6 +88,9 @@ function(xyz_add_test)
         CXX_STANDARD ${XYZ_VERSION}
         CXX_STANDARD_REQUIRED YES
         CXX_EXTENSIONS NO
+        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+        RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin
     )
 
     target_compile_options(${XYZ_NAME}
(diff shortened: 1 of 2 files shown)


cmake/xyz_add_test.cmake
```diff
@@ -101,7 +101,7 @@ function(xyz_add_test)
 
         include(GoogleTest)
         gtest_discover_tests(${XYZ_NAME}
-            WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+            WORKING_DIRECTORY $<TARGET_FILE_DIR:${XYZ_NAME}>
```
Twon, line 104: Actually, I think the correct fix would be to ensure the runtime output-dir is set per target as such:

```    set_target_properties(${XYZ_NAME}
        PROPERTIES
            ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
            LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin
    )
```

But if this works also happy with that for now

Twon (approved):

cmake/xyz_add_test.cmake
jbcoe: Happy to apply your fix but it does not seem to get rid of the error:

Is CMAKE_RUNTIME_OUTPUT_DIRECTORY the same as RUNTIME_OUTPUT_DIRECTORY?

Twon: Ok, that make sense as `RUNTIME_OUTPUT_DIRECTORY` is limited in scope to just the target where as `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is global in scope.  I think used in conjunction with your prior change will give us what we want, i.e. executables are in `build/bin`, without this they will be scattered throughout the build directory mirroring their offset from the project base to subdirectory.

PR #549, opened 2025-02-13
Avoid SFINAE in <=> using a lambda.

jbcoe:
Thanks to https://github.com/ned14 for the concept/SFINAE lesson.

### DIFF
--- a/indirect.h
+++ b/indirect.h
@@ -386,16 +386,6 @@ class indirect {
     return *lhs == *rhs;
   }
 
-  template <class U, class AA>
-  [[nodiscard]] friend constexpr auto operator<=>(const indirect<T, A>& lhs,
-                                                  const indirect<U, AA>& rhs)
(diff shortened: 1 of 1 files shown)


indirect.h
```diff
@@ -407,15 +397,31 @@ class indirect {
...
+
+  template <class U>
+  [[nodiscard]] friend constexpr auto operator<=>(const indirect<T, A>& lhs,
+                                                  const U& rhs) -> auto
+    requires(!is_indirect_v<U>)
+  {
+    // Define and call a lambda to allow requirements to be checked before
+    // the return type is determined.
+    // This avoids a recursive check inside __partially_ordered_with.
+    return [](const auto& lhs,
```
Twon, line 418: I don't think I understand this completely and want to!  How does defining the lambda like this allow for requirements to be checked before the return type is known?  

I would expect it still needs to be known only when the return type is...

Twon (approved):

PR #555, opened 2025-02-14
Add clang-tidy checks on CI

### LINKED ISSUE #550: Automate running of clang-tidy on CI
We'd like https://clang.llvm.org/extra/clang-tidy/ to be run as a GitHub code action.

jbcoe:
Add a clang-tidy CI action.

Add a script to create a duplicate-free version of compile_commands.json so that we don't have multiple entries for any source file.

Make `.cc` source non-PRIVATE in CMake libraries so that it gets mentioned in compile_commands.json (There may be a nicer fix for this).

Fixes #550

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -42,8 +42,8 @@ jobs:
       - name: clang-tidy
         run: |
             python scripts/reduce_compile_commands.py build/compile_commands.json > compile_commands.json
-            clang-tidy -warnings-as-errors="*" -p compile_commands.json \
-                indirect.cc \
-                indirect_test.cc \
-                polymorphic.cc \
-                polymorphic_test.cc
+            PROJECT_SOURCE_FILES="$(python scripts/reduce_compile_commands.py build/compile_commands.json \
+                --names-only \
+                --exclude_dirs build exploration compile_checks benchmarks)"
+            echo $PROJECT_SOURCE_FILES \
+              | xargs clang-tidy -warnings-as-errors="*" -p compile_commands.json
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -25,7 +25,7 @@ env:
   BUILD_TYPE: Debug
 
 jobs:
-  build:
+  clang-tidy:
     # The CMake configure and build commands are platform agnostic and should work equally
     # well on Windows or Mac.  You can convert this to a matrix build if you need
     # cross-platform coverage.
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -42,7 +42,7 @@ jobs:
       - name: clang-tidy
         run: |
             python scripts/reduce_compile_commands.py build/compile_commands.json > compile_commands.json
-            clang-tidy -p compile_commands.json \
+            clang-tidy -warnings-as-errors -p compile_commands.json \
                 indirect.cc \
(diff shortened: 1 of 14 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,49 @@
+name: Clang Tidy
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
+  push:
```
Twon, line 6: Do we need to worry about the push section, I assumed we had pushing to the main branch disabled?

CMakeLists.txt
```diff
@@ -55,7 +55,6 @@ xyz_add_library(
 target_sources(indirect_cxx14
     INTERFACE
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/indirect_cxx14.h>
-    PRIVATE
```
Twon, line 58: Do we want these files to be visible to users of this library? I would expect them to only care about the headers containing the implementation?

jbcoe: If I don’t put them here then they don’t show up in the compilation commands JSON file. 

I don’t know of a better fix, this works but is not ideal.

.github/workflows/clang-tidy.yml
jbcoe: Thanks. I’ll review the other actions too. They should be consistent.

CMakeLists.txt
Twon: I'm surprised that they would have to be declared as interface files to create a compile command.  The `INTERFACE` keyword in effect says they sound not be compiled within the indirect_cxx17 project, but should become part of the depending project.  

Is that the issue that you compile command for this file in a downstream target?

Twon: Oh I think I see, `xyz_add_library` creates an `INTERFACE` library so then does not create compile commands for it be cause the `INTERFACE` keyword explicitly says things from this target are needed for a downstream target only.

We could change the library to just be a static library (but this may cause linkage errors as there would be nothing to link) or move these to the test targets

jbcoe: My CMake taste is nowhere near as refined as yours. What would you prefer? 

Whatever we do needs an explanatory comment as it's weird. Having an empty source file is weird anyway so weird is perhaps to be expected.

Twon: Hmmm... what was the context for needing to compile these (that we did not get over doing so in the tests)?

If we don't really want it to be part of the library but want to compile then I would suggest were make a separate target per library which is an `OBJECT` library.  This is a library of compiled translation units.  Then we can keep them separate but still ensure we have a check which compiles each header (just they will become a separate target)

Does that make sense?  I can do this if that helps...

jbcoe: This sounds like the right fix - go team!

jbcoe: I've applied the fix @Twon suggested - worked easily.

.github/workflows/clang-tidy.yml
jbcoe: This is consistent with other CI checks. We should clean up (or not) in a follow-up PR.

CMakeLists.txt
Twon: Great, glad that has worked!

Twon (approved):
Looks good, thanks for applying the suggested changes.

jbcoe:
Thanks for helping me sort the CMake issues

PR #556, opened 2025-02-16
Add missing deduction guides and use `type_identity_t` for allocator extended copy/move ctors

jbcoe:
Enable more use of deduction guides.


BUILD.bazel
```diff
@@ -13,6 +13,7 @@ cc_library(
     srcs = ["indirect.cc"],
     hdrs = ["indirect.h"],
     copts = ["-Iexternal/value_types/"],
+    defines = ["XYZ_HAS_EXTENDED_CONSTRUCTOR_TEMPLATE_ARGUMENT_DEDUCTION"],
```
Quuxplusone, line 16: What is the purpose of `XYZ_HAS_EXTENDED_CONSTRUCTOR_TEMPLATE_ARGUMENT_DEDUCTION`? Does it cause problems for some compilers?
I would expect that all compilers should be able to support these deduction guides, and so you wouldn't need to hide them under a build flag.

feature_check.h
```diff
@@ -95,4 +95,18 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+//
+
+#ifdef XYZ_HAS_STD_TYPE_IDENTITY
+#error "XYZ_HAS_STD_TYPE_IDENTITY is already defined"
+#endif  // XYZ_HAS_STD_TYPE_IDENTITY
+
+#if (__cplusplus >= 202002L) || (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)
+#define XYZ_HAS_STD_TYPE_IDENTITY
+#endif  //(__cplusplus >= 202002L) || (defined(_MSVC_LANG) && _MSVC_LANG >=
+// 202002L)
```
Quuxplusone, line 110: This macro is set but never used. You use `type_identity_t` below without testing this macro.

indirect.h
```diff
@@ -465,10 +465,17 @@ concept is_hashable = requires(T t) { std::hash<T>{}(t); };
 template <typename Value>
 indirect(Value) -> indirect<Value>;
 
-template <typename Alloc, typename Value>
+template <typename Alloc, typename Value,
+          typename std::enable_if_t<!is_indirect_v<Value>, int> = 0>
```
Quuxplusone, line 469:
```suggestion
template <typename Alloc, typename Value>
```
The `enable_if` isn't needed, since the deduction guide on line 475 is more specific. (Please correct me if I'm wrong; I didn't test this.)

polymorphic.h
```diff
@@ -404,7 +414,19 @@ class polymorphic {
     }
   }
 };
-
+#ifdef XYZ_HAS_EXTENDED_CONSTRUCTOR_TEMPLATE_ARGUMENT_DEDUCTION
+template <typename Value>
+polymorphic(Value) -> polymorphic<Value>;
+
+template <typename Alloc, typename Value,
+          typename std::enable_if_t<!is_polymorphic_v<Value>, int> = 0>
```
Quuxplusone, line 422:
```suggestion
template <typename Alloc, typename Value>
```
Same here: the deduction guide on line 427 is more specific so the `enable_if` isn't needed.

And a big bonus here: you can get rid of your `xyz::is_polymorphic_v` which is confusingly dissimilar to `std::is_polymorphic_v`.

feature_check.h
jbcoe: It's used in tests to disable tests that won't run on older C++ versions.

polymorphic.h
jbcoe: `std::is_polymorphic_v<std::polymorphic<T>>` evaluating to `false` is a pretty compelling call for a rename. A new issue shall be raised.

Quuxplusone: FYI, `std::is_array_v<std::array<int, 10>>` and `std::is_function_v<std::function<int()>>` are both false too. (My [C++ Pub Quiz](https://quuxplusone.github.io/blog/2021/10/28/cppcon-2021-pub-quiz/) research is finally paying off! ;)) I think it's a _bonus_ to be able to _eliminate_ `xyz::is_polymorphic`, but if it's really indispensable, then I don't think you should worry about changing the name.

feature_check.h
Quuxplusone: Yeah, but it's not used to guard those features in `indirect.h`. So for example this just errors out:
```
clang++ -O2 -std=c++17 -c indirect_test.cc
```
Theoretically you could support CTAD in C++17 (because CTAD itself was a C++17 feature); but right now you have only your C++14 version (which can't use CTAD) and your C++20 version (which assumes full C++20 support).

I'm figuring all this out as I write, so, forgive me if this is all old news. I guess the README already implies that C++17 isn't a supported target. You can probably at least partially ignore this comment, then.

I'm still confused why `XYZ_HAS_EXTENDED_CONSTRUCTOR_TEMPLATE_ARGUMENT_DEDUCTION` is handled/set/tested in a different way from `XYZ_HAS_TEMPLATE_ARGUMENT_DEDUCTION`, though.

PR #558, opened 2025-02-21
An idea for the deduction guides

Quuxplusone:
Posting as a PR just so that the CI will run (I hope). This all works on my local machine.

I see now the problem with not-constraining the less specific deduction guide.
```
template <typename Alloc, typename Value>
polymorphic(std::allocator_arg_t, Alloc, Value) -> polymorphic<
    Value, typename std::allocator_traits<Alloc>::template rebind_alloc<Value>>;
```
Before we can do overload resolution between this deduction guide and the more specific one, we must fully evaluate this one's type... which requires evaluating `std::allocator_traits<Alloc>::template rebind_alloc<Value>`, which explodes because the deduced type `Alloc` _is not actually an allocator type._ So, even though the other deduction guide would have been the better match, this one explodes before we even get to the overload-resolution step.

Here's a reduced example of what I'm talking about: https://godbolt.org/z/jWzn5PGn6

It looks to me as if the STL currently doesn't try to rebind allocators in deduction guides at all, with the sole exception of `flat_set`/`flat_map`, which rebind _specifically_ in their `from_range_t` constructors and nowhere else. The `from_range_t` constructors (unlike the allocator-extended copy constructor) cannot deduce the allocator type in any other way, so it would be a hard error to pass in a non-allocator type there anyway.

So I think there would be good precedent for adopting an approach like the one in this PR. But I do recognize that this reduces the surface area of the library a little bit: I had to change some unit tests to avoid testing the pattern that will no longer work after this patch. However, notice that the analogous tests wouldn't pass for `std::vector` or `std::flat_set`, either!


jbcoe:
Thanks @Quuxplusone we don't want to be doing things differently to `vector` so this looks like a sound direction.

I've been recovering from a cold so will be slow to properly respond to this. 

On a first look, a minor nit is that we try to make indirect.h and polymorphic.h work as stand-alone single-include headers (for godbolt). I think your PR lightly couples them to features.h. It would, however, be easy to clean up post-merge.

PR #563, opened 2025-03-14
Move polymorphic control block definition into the class

jbcoe:


Twon (approved):
Agreed, the scope of this should be private

jbcoe:
> Agreed, the scope of this should be private

I’m not totally sure what you mean, should the rule itself be private?

Issue #559, opened 2025-02-28
Question: is this proof of concept reference implementation?


toughengineer:
In P3019 you refer to this repo as reference implementation.

If it is not a proof of concept and is somewhat production code, then e.g. polymorphic hierarchy `control_block`←`direct_control_block` in polymorphic.h is an absolute overkill and unnecessary overhead.

jbcoe:
This feels unnecessarily hostile.

We tried a few different approaches, benchmarks gave us no firm reason to favour one and the approach we’ve taken seems easy enough to follow.

What would your preferred implementation be?

toughengineer:
I did not mean to be hostile.

If you intend your code to be somewhat production grade and not just proof of concept then one would expect to get rid of unnecesary overhead.

Namely, considering example with `control_block` hierarchy from polymorphic.h, you never use `dynamic_cast` and the hierarchy is not exposed to the user, yet you impose overhead of compiler generating all the RTTI associated with the polymorphic hierarchy and linker linking it (and _hopefully_ eliminating the unused portion).

Instead I would suggest to use either pointer to tables of functions/operations like they do it for `std::any` in MSVC standard library, or use pointer to a function that does needed things (copy/move/destruction) depending on the operation code like they do it for `std::any` in libstdc++ and libc++.
In both cases you have a pointer to a table or to a function instead of (implicit) pointer to vtable, but you explicitly avoid generating all the unnecessary stuff related to `dynamic_cast` and other RTTI.

Also I would suggest to put pointer to the interface `T` (as in `polymorphic<T>`) inside the instance itself like it's done for `std::shared_ptr` instead of having it in the control block. This way you would not have extra indirection every time the user calls `operator->()` or `operator*()` which are basically the main way to use such an instance.
But that's a trade off between size of an instance and repeatedly (not) having an extra indirection, I guess.

Anyway, from you answer I conclude that you intend this to be somewhat production code and that answers my question.

PR #567, opened 2025-04-14
Replaced inheritance and virtual functions with a pointer to function

toughengineer:
Following discussion in #559 I replaced inheritance and virtual functions with a single pointer to function.

With the current benchmarks I could not reliably measure the difference in performance so I assumed that it is not significantly different.

I could however very easily measure the difference in resulting binary size.
Here's the size of **polymorphic_test.exe** in KB:

|compiler|current size|size after change|difference|
|-|-|-|-|
|MSVC|921|904|-17|
|Clang|996|982|-14|

I did not investigate what exactly contributed to overhead, I bet it's RTTI, and maybe vtables and individual functions a little bit.

Solution with a pointer to a single function (apart from binary size compared to inheritance and virtual functions) is very ABI friendly, and can be made extendable in an ABI compatible way.

These are roughly the minimal changes that are needed, I tried not to touch parts that are irrelevant to this particular problem, i.e. I did not refactor the rest of the code.

Please feel free to benchmark it yourself and/or modify the code to your liking.

### DIFF
--- a/polymorphic.h
+++ b/polymorphic.h
@@ -53,10 +53,18 @@ class polymorphic {
     using allocator_traits = std::allocator_traits<A>;
     typename allocator_traits::pointer p_;
 
-    virtual constexpr ~control_block() = default;
-    virtual constexpr void destroy(A& alloc) = 0;
-    virtual constexpr control_block* clone(const A& alloc) = 0;
-    virtual constexpr control_block* move(const A& alloc) = 0;
+    enum class Action { Destroy, Clone, Move };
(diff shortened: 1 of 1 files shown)


jbcoe:
Thanks so much for the work on this. I'm inclined to adapt the work and add another implementation rather than replace the one we have. Unlike indirect, I think there's significant implementation freedom with polymorphic and lots of space to explore.

I'll work on adapting this PR to introduce a parallel implementation. No input needed.

jbcoe:
This PR can be replaced by #571

Issue #626, opened 2026-02-25
Suggestion: Package indirect.h and polymorphic.h inside a `xyz_value_types` directory


jogerh:
When installing value_types (cmake install or vcpkg), the header files `indirect.h` and `polymorphic.h` end up in the top level include directory. It could be better to group these inside an xyz_value_types subdirectory, so that clients would include them like `#include <xyz_value_types/indirect.h>` instead of `#include <indirect.h>`. This is mostly a cosmetic topic, but it seems to be quite common practice for other libraries.

jbcoe:
I'd recommend using:

* https://github.com/bemanproject/indirect

for (likely) higher quality implementations with ongoing support (and vcpkg).

PR #627, opened 2026-03-15
Move files so that includes come from xyz/value_types/

jbcoe:
Addresses #626 

Manual work is needed to test packages after this change.


philipcraig (approved):
LGTM

jbcoe:
This is stale. Easier to re-do than re-base.

